PR #2, opened 2015-11-21
Use logger for error reporting

ppolewicz:
http://inventwithpython.com/blog/2012/04/06/stop-using-print-for-debugging-a-5-minute-quickstart-guide-to-pythons-logging-module/

However I am not entirely sure if logging is a good idea here. Therefore please review before merging.

### DIFF
--- a/b2
+++ b/b2
@@ -24,6 +24,11 @@ import urllib
 import urllib2
 import base64
 
+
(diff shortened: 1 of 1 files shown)


ppolewicz:
It was a bad idea.

bwbeach:
Thanks for the update.  :-)

PR #3, opened 2015-11-30
Refactor and optimize StoredAccountInfo

ppolewicz:


bwbeach:
Thanks for the changes.  I wanted most of them, but not all, so I just re-applied them to my branch.

ppolewicz:
@bwbeach why did you drop `refresh_entire_bucket_name_cache()`?

As for logging, you could rebase it and drop a portion you didn't want instead of doing it all again manually. Or you could just ask me to do it and update the branch related to this pull request, so you'd just need to click "merge".

bwbeach:
With only 100 buckets per account, I didn't care about the optimization of `refresh_entire_bucket_name_cache`.

Someday I'll learn how to rebase in git.

ppolewicz:
Right, I didn't know about the 100 buckets limit - I thought there is a limit of 2500 free bucket creations per day.
I'd rather not rewrite the config 100 times if it can be avoided (it can hurt people that have $HOME on NFS for example), but if the bucket names won't change remotely too much, then maybe it will not be a problem (current pricing model discourages such usage ;) ).

If you want to learn how to use the power of git, here is a great free book that I can recommend: https://git-scm.com/book/en/v2

bwbeach:
That's a good point about rewriting the file 100 times.  That's a good enough reason to make the change.

ppolewicz:
Then I will make a new pull request for including that optimization tonight.

PR #10, opened 2015-12-12
Fix write mode

ferricoxide:
Fixes regression from PR #6. Using truncating, rather than appending, open operation.

This update will not correct the permission on `${HOME}/.b2_account_info` if it already exists and is already incorrectly set.


ppolewicz:
This fix is incomplete. Please use #9 instead.

bwbeach:
Took ppolewicz's change instead.

PR #21, opened 2015-12-26
Add a progress bar

ppolewicz:


ppolewicz:
@bwbeach if there is a pep8 violation (or some other problem), please raise it in review and ask me to fix it rather than fix it yourself. This way I can learn (and you can scale much better)!

bwbeach:
Will do.  Thanks.

I've started using the free, community edition of PyCharm, which includes PEP 8 warnings.

ppolewicz:
If you don't mind some CI, I'll add automatic verification of such things for this project after I complete #11.

bwbeach:
Sounds good.

I added a test program, but it needs account credentials to run, and I didn't want to make those public.

Issue #23, opened 2015-12-31
Need --quiet option to make output easy to parse.


bwbeach:
Some people have written scripts around `b2` that parse the output  I plan on adding a `--quiet` option for uploads and downloads to disable the progress bar.  We can add the same option everywhere later.

ppolewicz:
Actually, progress bar is emitted to stderr, so it will not affect whoever is parsing the output, unless they intercept stderr explicitly.

If there is a problem, we could report it on stdout and return an error code specific to the problem, as other console commands do (`curl`, for example).

Nevertheless, `--quiet` sounds like a good idea.

After #11 we can refactor the commands and option parsing very neatly, so that adding such options will become very easy.

bwbeach:
Some people check stderr, and treat any output there as an error.

I'll just hack in --quiet on uploads and downloads, to preserve current behavior for now.  Then we can add it generally when arg parsing gets redone.

bwbeach:
Pull request merged.

ppolewicz:
@bwbeach support for hidden global flags has been added in 46ea7456157cc9b9abfe86670ad1f0ecc1fd5da1 yesterday. As far as I know, there is no strong demand to implement #23. Perhaps it would be better to close it and reopen once the demand returns?

PR #24, opened 2015-12-31
Add --quiet option to upload_file.

bwbeach:
This allows scripts to easily parse the JSON that upload_file prints to stdout.

### DIFF
--- a/b2
+++ b/b2
@@ -652,14 +652,15 @@ def list_file_versions(args):
     print json.dumps(response, indent=2, sort_keys=True)
 
 
-def ensure_upload_data(bucket_id, info):
+def ensure_upload_data(bucket_id, info, quiet):
     """
     Makes sure that we have an upload URL and auth token for the given bucket and
     returns it.
     """
     upload_data = info.get_bucket_upload_data(bucket_id)
     if upload_data is None:
-        print 'Getting upload URL...'
+        if not quiet:
+            print 'Getting upload URL...'
         auth_token = info.get_account_auth_token()
         url = url_for_api(info, 'b2_get_upload_url')
         params = { 'bucketId' : bucket_id }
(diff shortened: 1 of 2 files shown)


b2
```diff
@@ -801,9 +806,9 @@ def upload_file(args):
             headers['X-Bz-Info-' + k] = b2_url_encode(v)
 
         try:
-            response = post_file(url, headers, local_file, exit_on_error=False, progress_bar=True)
+            response = post_file(url, headers, local_file, exit_on_error=False, progress_bar=(not quiet))
```
ppolewicz: The extra parenthesis look weird. I would make a new boolean variable `progress_bar` before the `try:` statement to make it easier to read. It doesn't matter very much, we'll clean it up next week anyway.

b2
```diff
@@ -801,9 +806,9 @@ def upload_file(args):
...
             print json.dumps(response, indent=4, sort_keys=True)
-            if 'fileId' in response:
+            if (not quiet) and ('fileId' in response):
```
ppolewicz:
```python
if 'fileId' in response and not quiet:
```
has much less parenthesis and is much more readable (reads like english)

ppolewicz:
The commit should have "Closes #23" or "Resolves #23" in the message so that #23 is closed automatically upon merging.

bwbeach:
I'll do the "Closes #23" next time.

b2
bwbeach: As you can tell, I tend to prefer more parens when there might be any question about operator precedence.  This case could go either way.

ppolewicz: True, but I find the suggested way to be more pythonic (and more readable to python coders) than the one you used, that is why I suggested it.

ppolewicz:
In such case the last commit message can be simply changed by running `git commit --amend` and pushing the topic branch with `-f` flag, so `git push -f origin quiet_uploads`. (please don't push the master branch with `-f` or forks may break for other people).

PR #28, opened 2016-01-06
Preparation for downloading refactor

ppolewicz:


ppolewicz:
@bwbeach tests pass, ready to merge

bwbeach:
Nice work.   I'll go ahead and merge it.

My only complaint is that I much prefer "while True" to "while 1", even though the latter may be a tiny bit faster.

ppolewicz:
I've seen enough high performance code to get used to it. Unfortunately `while True` in python is slow... But how about this then:
```python
with stream as f:
    data = response.read(block_size)
    while data:
        f.write(data)
        digest.update(data)
        bytes_read += len(data)
        data = response.read(block_size)

if set_last_modified:
```
It is more pythonic than the current and previous solution. It should also be faster than the current version due to less checks. Would it be better?

bwbeach:
Slow is a relative term.  The loop has network I/O in it, which is going to be significantly slower than evaluating True.

And I would choose "while 1" over duplicating the call to read().

My preference for "while True" over "while 1" comes from programming in languages, like Java, that have a separate type for boolean, and don't automatically convert other types to boolean.  And then there's PHP, where the conversion-to-boolean rules are just crazy, and you want to avoid them at all cost.

Anyway, it's not a big deal.  I just wanted to let you know what my style preferences are.  Your code is perfectly readable, and I'd rather focus or time on adding new features.

ppolewicz:
If I write in Java, my preference is the same as yours.

As much as I hate python for needing funny tricks like `temp_variable = self.child.grandchild.field` before a loop (because the "dot" is expensive due to possible `__gettattr__`) and `while 1:` loops, it happens to be "the way to go" in python. While I refrain myself from optimizing all code which I see to be as efficient as possible, I do tweak it if it gets in my way on some unrelated occasion (unless it degrades readability severely). I hope you don't mind.

That being said, I know I am going to "slow down" `ls` by introducing an API layer which returns objects instead of printing strings on a terminal, but the object interface is more important than performance of writing strings on the terminal.

PR #32, opened 2016-01-07
tqdm inheritance fix

ppolewicz:
Also add a suppression for flake8 to ignore pyflakes alert about redefinition of imported module


ppolewicz:
I'll work on `ls` next.

PR #41, opened 2016-01-15
Implement AbstractAccountInfo.authorize() and ConsoleTool.authorize_account()

ppolewicz:


bwbeach:
I tend to prefer delegation over inheritance.  I could imagine the stored account info delegating to an object that knows how to do the authentication.

Another possibility is what we did in the Java libraries internal to Backblaze: one class that knows how to make the raw API calls, and a layer on top of it that knows how to keep track of the auth tokens, upload URLS, etc.

Meanwhile, this change looks OK.

ppolewicz:
If in the future there will be possibility to use various authentication protocols, it will be important to separate the protocol from the information storage layer. Then it will be much better to use delegation. But it doesn't look like its coming and meanwhile I had to put it *somewhere*.

PR #42, opened 2016-01-16
Raw api layer

bwbeach:
This is a proposal for a "raw API" layer underneath the B2Api class.  The idea is to have a super-simple HTTP wrapper class that doesn't need much testing itself, and can be replaced with a mock or simulator class for testing the layers of code above it.

### DIFF
--- a/b2
+++ b/b2
@@ -763,10 +763,75 @@ class AuthInfoCache(AbstractCache):
         self.info.refresh_entire_bucket_name_cache(self._name_id_iterator(buckets))
 
 
+## B2RawApi
+
+
+class B2RawApi(object):
+
+    """
+    Provides access to the B2 web APIs, exactly as they are provided by B2.
+
+    Requires that you provide all necessary URLs and auth tokens for each call.
+
+    Each API call decodes the returned JSON and returns a dict.
+
+    For details on what each method does, see the B2 docs:
+        https://www.backblaze.com/b2/docs/
+
+    This class is intended to be a super-simple, very thin layer on top
+    of the HTTP calls.  It can be mocked-out for testing higher layers.
+    And this class can be tested by exercising each call just once,
+    which is relatively quick.
+    """
+
+    def authorize_account(self, realm_url, account_id, application_key):
+        auth = 'Basic '+ base64.b64encode('%s:%s' % (account_id, application_key))
+        url = self._build_url(realm_url, 'b2_authorize_account')
+        return post_json(url, {}, auth)
+
+    def create_bucket(self, api_url, account_id, account_auth_token, bucket_name, bucket_type):
+        url = self._build_url(api_url, 'b2_create_bucket')
+        params = {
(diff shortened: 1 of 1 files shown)


b2
```diff
@@ -763,10 +763,75 @@ class AuthInfoCache(AbstractCache):
...
+    buckets and files returned are Python objects with accessor
+    methods.
+
+    Also,  keeps a cache of information needed to access the service,
+    such as auth tokens and upload URLs.
+    """
+
+    # TODO: move HTTP code out to B2RawApi
+    # TODO: ConsoleTool passes the account info cache into the constructor
+    # TODO: provide method to get the account info cache (so ConsoleTool can save it)
```
ppolewicz: `B2Api.cache` is public and by default it uses StoredAccountInfo or wrapper over it, which saves the cache in `~/.b2_account_info`. Other (other than `ConsoleTool`) `B2Api` users are likely to use the same capability to just `api = B2Api()` and call methods on it, if they use a single thread.

b2
```diff
@@ -1327,6 +1390,17 @@ def download_file_by_name(args):
 
 
 class ConsoleTool(object):
+
+    """
+    Implements the commands available in the B2 command-line tool
+    using the B2Api library.
+
+    Responsible for storing cached account data in ~/.b2_account_info,
```
ppolewicz: I don't think `ConsoleTool` should be responsible for that - other `B2Api` callers should be saved from doing it in their code unless it is necessary.

ppolewicz:
I am a bit torn on this. We can mock urllib for tests as it currently is and splitting raw api from the public API adds a new layer, which makes working with the code slightly more complicated. It allows for more fine-grained testing, which we may not need at this point.

It will be a lot of work.

It will conflict with everything.

Lets do it anyway!

b2
bwbeach: I think that's true, and it is fine as a default.  There may be users of the library, though, that don't want to use `~/.b2_account_info`, and want to use a different kind of cache.  They may want to provide the account ID and application key when they construct a B2Api, and then cache based on that.

b2
bwbeach: How about:

> Uses the StoredAccountInfo object to keep account data in `.b2_account_info` between runs.

bwbeach:
Will this conflict with any work you have in progress?  I'm happy to wait until an opportune time, and merge my changes then.

b2
ppolewicz: Perfect

ppolewicz:
It will conflict with nearly any change in `B2Api`. Please merge it as it currently is and lets sync when someone works on what so that we can avoid bigger conflicts.

b2
ppolewicz: Indeed. They can pass `account_info` and `cache` arguments to `B2Api` for exactly that purpose. `InMemoryCache` is implemented already, but I haven't gotten around to `InMemoryAccountInfo` yet.

PR #48, opened 2016-01-26
Allow credential file name to be overridden

Papafox:
Running b2 under a low privileged user, such as user nobody, has problems with the credential file.

Typically uses such as these do not have a home directory which means that '~/.b2_account_info' will resolve to '/.b2_account_info'.  This in turn means that the user will require write access to the root directory which is bad.

The solution is to allow the name of the credential file to be overridden using an environment variable 'B2_ACCOUNT_INFO'.  If this variable does not exist, the name will default to '~/.b2_account_info'

Example:
export B2_ACCOUNT_INFO=/tmp/b2_credentials
b2 authorize_account ${ACCT_ID} ${ACCT_KEY}

Thus B2 will store its credentials in the file '/tmp/b2_credentials' rather than the root directory

### DIFF
--- a/b2
+++ b/b2
@@ -25,7 +25,7 @@ import getpass
 import hashlib
 import httplib
 import json
-import os.path
+import os.path, os.environ
 import socket
 import stat
 import sys
--- a/b2
+++ b/b2
@@ -1113,9 +1113,8 @@ class StoredAccountInfo(AbstractAccountInfo):
     DOWNLOAD_URL = 'download_url'
 
     def __init__(self):
-        account_info_filename = os.environ['B2_ACCOUNT_INFO']
-        if account_info_filename is None : account_info_filename = '~/.b2_account_info'
-        self.filename = os.path.expanduser(account_info_filename)
+        user_account_info_path = os.environ.get('B2_ACCOUNT_INFO', '~/.b2_account_info')
+        self.filename = os.path.expanduser(user_account_info_path)
         self.data = self._try_to_read_file()
         if self.BUCKET_UPLOAD_DATA not in self.data:
             self.data[self.BUCKET_UPLOAD_DATA] = {}
--- a/b2
+++ b/b2
@@ -59,7 +59,8 @@ Usages:
         The application key is a 40-digit hex number that you can get from
         your account page on backblaze.com.
 
-        Stores an account auth token in ~/.b2_account_info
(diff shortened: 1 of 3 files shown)


b2
```diff
@@ -25,7 +25,7 @@ import getpass
 import hashlib
 import httplib
 import json
-import os.path
+import os.path, os.environ
```
ppolewicz: Please keep the imports ordered alphabetically.

Also that import fails in our CI environment. Which version of python have you tested it on?

b2
```diff
@@ -1113,7 +1113,9 @@ class StoredAccountInfo(AbstractAccountInfo):
     DOWNLOAD_URL = 'download_url'
 
     def __init__(self):
-        self.filename = os.path.expanduser('~/.b2_account_info')
+        account_info_filename = os.environ['B2_ACCOUNT_INFO']
```
ppolewicz: If `B2_ACCOUNT_INFO` is not exported, this will raise an exception. Maybe it could be done like this:

```python
user_account_info_path = os.environ.get('B2_ACCOUNT_INFO')
self.filename = user_account_info_path or os.path.expanduser('~/.b2_account_info')
```

?

ppolewicz:
@Papafox this seems like a good idea and I support it fully, however there are a few issues with your implementation. See the Travis CI build status which checks style and imports (click the "Details" below last comment). We have a pre-commit script placed in the repository to check such things locally before commiting.
Also please consider making a test for the functionality which you have implemented, so that we are sure it will not be broken when performing future work on b2 command line tool.

Papafox:
I've been testing with python v2.7.1.  I don't understand why importing os.environ fails - widening the import to plain os fixes it anyway.

Wait another 24 hours and I'll write a test case.  I agree test cases are always good.

ppolewicz:
Cool, thanks.

By the way, please see my other review comment (in "Files changed" tab)

Papafox:
I've added a test case to the basic test suite.  It renames the existing credential file and copies it as a new credential file.  An environment variable B2_ACCOUNT_INFO is created, which points at the new credential file.

Three tests are then run:
* clear_account to clear the credential file
* authorize_account with bad credentials
* authorize_account with good credentials

The test environment is then torn-down with the previous credential file being renamed back to it's original name, the new credential being deleted and the environment variable being unset.

Both the documentation (README.md) and the inline help have been updated to mention the new environment variable.

PR #53, opened 2016-01-31
More groundwork for unit tests.

bwbeach:
I'd like to re-do the way the sync command works, but want to have unit tests before doing so.  This will mean moving the rest of the HTTP calls into B2RawApi, and then building a stub class to simulate B2RawApi.

This initial change moves the bucket-related calls into B2RawApi.

### DIFF
--- a/b2
+++ b/b2
@@ -415,8 +415,9 @@ class Bucket(object):
     MAX_UPLOAD_ATTEMPTS = 5
     MAX_UPLOADED_FILE_SIZE = 5 * 1000 * 1000 * 1000
 
-    def __init__(self, api, id_, name=None, type_=None):
-        self.api = api
+    def __init__(self, raw_api, account_info, id_, name=None, type_=None):
+        self.raw_api = raw_api
+        self.account_info = account_info
         self.id_ = id_
         self.name = name
         self.type_ = type_
@@ -425,14 +426,12 @@ class Bucket(object):
         return self.id_
 
     def set_type(self, type_):
-        account_info = self.api.account_info
-        auth_token = account_info.get_account_auth_token()
-        account_id = account_info.get_account_id()
+        auth_token = self.account_info.get_account_auth_token()
+        account_id = self.account_info.get_account_id()
 
-        url = url_for_api(account_info, 'b2_update_bucket')
-        params = {'accountId': account_id, 'bucketId': self.id_, 'bucketType': type_,}
-        response = post_json(url, params, auth_token)
-        return response
+        return self.raw_api.update_bucket(
+            self.account_info.get_api_url(), auth_token, account_id, self.id_, type_
+        )
 
     def ls(self, folder_to_list='', show_versions=False, max_entries=None, recursive=False):
         """Pretends that folders exist, and yields the information about the files in a folder.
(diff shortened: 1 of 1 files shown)


b2
```diff
@@ -840,11 +843,29 @@ class B2RawApi(object):
         url = self._build_url(realm_url, 'b2_authorize_account')
         return post_json(url, {}, auth)
 
-    def create_bucket(self, api_url, account_id, account_auth_token, bucket_name, bucket_type):
+    def create_bucket(self, api_url, account_auth_token, account_id, bucket_name, bucket_type):
         url = self._build_url(api_url, 'b2_create_bucket')
-        params = {'accountId': account_id, 'bucketName': bucket_name, 'bucketType': bucket_type,}
+        params = {'accountId': account_id, 'bucketName': bucket_name, 'bucketType': bucket_type}
+        print url
```
ppolewicz: leftover prints

b2
```diff
@@ -840,11 +843,29 @@ class B2RawApi(object):
...
+        print params
         return post_json(url, params, account_auth_token)
 
+    def delete_bucket(self, api_url, account_auth_token, account_id, bucket_id):
+        url = self._build_url(api_url, 'b2_delete_bucket')
+        params = {'accountId': account_id, 'bucketID': bucket_id}
+        return post_json(url, params, account_auth_token)
+
+    def list_buckets(self, api_url, account_auth_token, account_id):
```
ppolewicz: this could be refactored with `delete_bucket` and `update_bucket` which are almost identical

b2
```diff
@@ -895,9 +916,10 @@ class B2Api(object):
         auth_token = self.account_info.get_account_auth_token()
 
         response = self.raw_api.create_bucket(
-            self.account_info.get_api_url(), account_id, auth_token, name, type_
+            self.account_info.get_api_url(), auth_token, account_id, name, type_
         )
-        bucket = BucketFactory.from_api_bucket_dict(self, response)
+        print response
```
ppolewicz: leftover print

ppolewicz:
There are two changes glued together in a single commit: moving of some http methods to B2RawApi and a change in how Bucket objects are constructed.

The second change forces the user to explicitly provide an instance of raw_api and account_info when constructing the bucket object. It is a non-small change in the API interface (thank you for keeping it on a branch for discussion).

It used to be
```py
api = B2Api()
bucket_dict = api.something('...')
bucket = BucketFactory.from_api_bucket_dict(api, bucket_dict)
```
and now it will be
```py
api = B2Api()
bucket_dict = api.something('...')
raw_api = B2RawApi()
account_info = StoredAccountInfo()  # wrong! should be account_info = B2Api().account_info
bucket = BucketFactory.from_api_bucket_dict(raw_api, account_info, bucket_dict)
```

The change also isolates the Bucket class from the B2Api class completely, which might be fine, but we might want to use some B2Api methods from something like `Bucket.sync_down()` (for example `B2Api.download_by_id()`). Passing a B2Api instance encapsulated its properties like `account_info`, `raw_api`, `cache` or any other (maybe `b2_session` in the future), which might just be useful when performing all kinds of bucket operations. Therefore I think we should not make this change and keep Bucket having an `api` as a member. At the same time, it is ok to do something like `self.raw_api = api.raw_api` in `Bucket.__init__` for convenience.

As for moving the actual http calls from Bucket to B2RawApi, I wholeheartedly approve.

PR #54, opened 2016-01-31
pre-commit.sh now displays diffs when updating the formatting of the …

bwbeach:
…code,

and script won't quit in the middle if code needs updating.

Also, added friendly messages when checks pass, so it's not just silent.


ppolewicz:
I approve

PR #55, opened 2016-02-01
Second try at moving bucket-related calls to B2RawApi.

bwbeach:
Thanks to Pawel for the useful comments.


ppolewicz:
This change is very neat indeed. I approve!

PR #58, opened 2016-02-11
Adding options to clear bucket of all files

skeetio:
Added an option to clear a bucket of all files.  Added a user prompt that they must verify before it continues and an optional argument to disable this prompt.

### DIFF
--- a/b2
+++ b/b2
@@ -647,6 +647,36 @@ class Bucket(object):
             result['bucketType'] = self.type_
         return result
 
+    def clear_bucket(self):
+        last_file = None
+        api_responses = []
+
+        while True:
+            response = self.list_file_names(start_filename=last_file, max_entries=100)
+
+            if len(response['files']) == 0:
+                break
+
+            try:
+                del_files = tqdm(
+                    response['files'],
+                    desc='Deleting next batch of %s' % len(response['files'])
+                )
+                has_progress_bar = True
+            except TypeError:
+                has_progress_bar = False
+                del_files = response['files']
+
+            for file_data in del_files:
+                self.api.delete_file_version(file_data['fileId'], file_data['fileName'])
+                if has_progress_bar is False:
+                    print '- %s' % file_data['fileName']
+
+            last_file = response['nextFileName']
+            api_responses.append(response)
+
+        return api_responses
+
     def __repr__(self):
         return 'Bucket<%s,%s,%s>' % (self.id_, self.name, self.type_)
 
@@ -1527,20 +1557,8 @@ class ConsoleTool(object):
                 return
 
         bucket = self.api.get_bucket_by_name(bucket_name)
-        last_file = None
 
-        while True:
-            response = bucket.list_file_names(start_filename=last_file, max_entries=100)
-
-            if len(response['files']) == 0:
-                break
-
-            for file_data in response['files']:
-                self.delete_file_version([file_data['fileName'], file_data['fileId']])
-
-            last_file = response['nextFileName']
-
-        print '%s cleared of all files' % bucket_name
+        bucket.clear_bucket()
 
     def delete_bucket(self, args):
         if len(args) != 1:
--- a/b2
+++ b/b2
@@ -1518,7 +1518,9 @@ class ConsoleTool(object):
         bucket_name = args[0]
 
         while force_delete is False:
-            user_input = raw_input('Are you sure you want to delete all the files in this bucket? (Y/N) ').upper()
+            user_input = raw_input(
+                'Are you sure you want to delete all the files in this bucket? (Y/N) '
+            ).upper()
             if user_input in ['Y', 'YES']:
                 break
             elif user_input in ['N', 'NO']:
--- a/b2
+++ b/b2
@@ -66,6 +66,12 @@ Usages:
 
         Erases everything in ~/.b2_account_info
 
+    b2 clear_bucket [--force] <bucketName>
+
+        Deletes all files in the named bucket.  The force argument will
+        suppress a prompt asking if the user is sure they wish to delete
+        all files.
+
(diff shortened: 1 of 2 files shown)


b2
```diff
@@ -1499,6 +1505,43 @@ class ConsoleTool(object):
...
+        while force_delete is False:
+            user_input = raw_input(
+                'Are you sure you want to delete all the files in this bucket? (Y/N) '
+            ).upper()
+            if user_input in ['Y', 'YES']:
+                break
+            elif user_input in ['N', 'NO']:
+                return
+
+        bucket = self.api.get_bucket_by_name(bucket_name)
```
ppolewicz: This should be moved to the Bucket class

skeetio:
As you wish.  Moved the clear bucket logic into the Bucket class and added some user feedback using tqdm if available or just a simple print of the file name as it is being deleted.

b2
```diff
@@ -641,6 +647,36 @@ class Bucket(object):
...
         return result
 
+    def clear_bucket(self):
+        last_file = None
+        api_responses = []
+
+        while True:
+            response = self.list_file_names(start_filename=last_file, max_entries=100)
+
+            if len(response['files']) == 0:
```
ppolewicz:
```py
if not response['files']:
```

ppolewicz:
@skeetio thank you for your effort, it is really appreciated.

I noticed that after the progress bar change, API is using tqdm or print. The API itself is designed to be used in other programs, like FUSE filesystem for example, or a GUI tool, web page or something.

In the past every method handled errors with `sys.exit(1)`, which would terminate any of the above.

For this reason, we isolated the user interface from the actual logic, so that `ConsoleTool` can be used in case the tool is invoked from the shell and in other case the API and Bucket classes provide all input and output other interfaces could use.

Now, after the last change, this separation has been compromised, which would result in unwanted behavior in any application (other than `ConsoleTool`).

Would you be willing to work with me to improve it further, so that the progress bar functionality of `Bucket.clear()` and isolation of user interface are both preserved?

skeetio:
Would it be acceptable to implement this similar to upload_file with an optional parameter that defaults to suppressing printed/tqdm output?

ppolewicz:
"Acceptable" is a very good word to describe this solution. Ideally we would pass a progress reporter object, which would default to `None`, which would then automatically create a dummy object which doesn't do anything, and use this one. In case of `ConsoleTool`, a smarter progress reporter (based on `tqdm` or simple prints) would be used.
Then, in case of GUI application or web interface, some other progress reporter implementation could be plugged to report the progress. Current `upload_file` implementation is a kind of a hack that I hope we will eradicate (using the approach described above), so it would be best not to clone this approach further. If you don't have resources to implement it in a fully flexible way, the `quiet` flag will have to do for now and we will change it later.

skeetio:
Let me see what I can do.  Would it be your preference that the progress object be passed at the class level to be used by any methods as they see fit or to individual methods on an as-needed basis?

ppolewicz:
We must ask ourselves a question. If we were using this code as a library in our program, would we rather pass the progress reporter object to the `B2Api` constructor, or would we rather want to pass it to the methods? (this makes me lean toward passing it once)

What about concurrency? If we would pass a single progress reported object to `B2Api`, it would receive progress reports in an interleaved manner for each thread/greenlet running in parallel. (this makes me lean toward passing it in each method).

Maybe for the case of threading, we should pass a progress reporter factory into B2Api, which would then create a progress reporter for each file? That becomes complicated. I don't yet have a good solution in mind for this.

bwbeach:
I'm glad you're thinking about concurrency.  

I think I would favor passing the progress reporter to the B2Api constructor, assuming that we define a thread-safe progress reporting interface.  Maybe the B2Api constructor should get a progress reporter factory, so that it can use a separate progress reporter object for each action.

ppolewicz:
@skeetio We seem to have a similar idea with Brian about progress reporter factory (passed to B2Api constructor). It would be best approach considering future features.

bwbeach:
On the other hand, passing a progress reporter to each method in B2Api would let the application code pick the title string for each reporter.  If B2Api is calling a factory, it will either have to provide the title string for the progress reporter, or tell the factory what it's about to do (e.g., upload a file).  Passing a progress reporter into the upload_file method would keep B2Api simpler.  I think I've changed my mind, and now prefer this approach.

ppolewicz:
Now that you mention it, I noticed that passing state between progress reporter factory and api caller would be difficult. It would be much easier to pass it explicitly to upload_file etc

skeetio:
I feel like the scope of this has progressed far beyond what the original intent of the PR was, which was to add a mechanism to clear out a bucket, which is a prerequisite for bucket deletion. From the perspective of a user who had buckets in excess of 500+ files there was no good way to completely wipe and delete a bucket (including the web ui, which had timeout issues with deletes as small as 30 files).

How about we strip out the progress reporting aspect of the clear bucket method and tackle a more robust reporter implementation in a separate pr/branch?

ppolewicz:
I approve. Tbh I was surprised to see progress reporting in the same PR as clear_bucket.

ppolewicz:
By the way, would you consider adding a test for clear_bucket?

skeetio:
Will do. Also will add the method to the bash completions

ppolewicz:
Hi @skeetio,

I noticed that some time has passed in this issue. Do you intend to finish this?

bwbeach:
No action for a long time.  Closing.

nilayp:
Just wanted to know there is a "workaround" to clearing a bucket of all it's files using B2 CLI's sync command. It is documented on the Backblaze B2 knowledge base: https://help.backblaze.com/hc/en-us/articles/225556127-How-Can-I-Easily-Delete-All-Files-in-a-Bucket-

PR #63, opened 2016-02-17
Add a custom user-agent string.

bwbeach:
At Backblaze, we'd like to know which requests are coming from
this tool/library.


snorrelo:
It would be nice if the user-agent string used the standard format. http://tools.ietf.org/html/rfc7231#section-5.5.3

bwbeach:
Thanks for the info on user-agent format.  I'll update the code to match the standard.

PR #64, opened 2016-02-17
BUG FIX: b2 sync on Windows 

HenrikBengtsson:
`b2 sync` gave server error on `file names must not contain '\\'` because pathnames on Windows uses backslashes which are kept as is when constructing the corresponding remote pathnames.


bwbeach:
Your version number change conflicted with mine.  I'll merge the fix in.

PR #66, opened 2016-02-18
Bash completion usage

ppolewicz:
I tried to set up bash completion using the previous guide but it was insufficient. I am not an expert on this subject, but hopefully with this information it is easier for otherts to install it.


bwbeach:
Thanks.  This will be very helpful.

PR #75, opened 2016-03-02
Workarounds for Python 3 json/str/bytes issues.

bwbeach:
json.load() doesn't work on http responses in
python 3.  this change adds explicit encoding
and decoding where needed.


ppolewicz:
When is it exposed? Why this has not been detected in tests?

bwbeach:
Summary of conversation on github:  

- tests now pass in 2.7, 3.4, and 3.5
- will merge and release, then investigate further

PR #79, opened 2016-03-06
Finish moving all API calls into B2RawApi.

bwbeach:
This is in preparation for unit tests.  The next step will be to build a stub/simulator for B2RawApi that the unit tests can use.

### DIFF
--- a/b2/b2.py
+++ b/b2/b2.py
@@ -865,6 +865,15 @@ class B2RawApi(object):
             bucketId=bucket_id
         )
 
+    def delete_file_version(self, api_url, account_auth_token, file_id, file_name):
+        return self._post_json(
+            api_url,
+            'b2_delete_file_version',
+            account_auth_token,
+            fileId=file_id,
+            fileName=file_name
+        )
+
     def get_upload_url(self, api_url, account_auth_token, bucket_id):
         return self._post_json(api_url, 'b2_get_upload_url', account_auth_token, bucketId=bucket_id)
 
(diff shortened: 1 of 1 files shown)


b2/b2.py
```diff
@@ -467,18 +467,21 @@ def ls(self, folder_to_list='', show_versions=False, max_entries=None, recursive
...
-        url = url_for_api(self.api.account_info, api_name)
         start_file_name = prefix
         start_file_id = None
         while True:
             params = {'bucketId': self.id_, 'startFileName': start_file_name}
             if start_file_id is not None:
                 params['startFileId'] = start_file_id
-            response = post_json(url, params, auth_token)
+            raw_api = self.api.raw_api
+            api_url = self.api.account_info.get_api_url()
```
ppolewicz: after this change, `api_url` and `auth_token` are refreshed on every iteration of the loop. This will affect performance negatively.

b2/b2.py
```diff
@@ -969,6 +1028,13 @@ def delete_bucket(self, bucket):
             self.account_info.get_api_url(), auth_token, account_id, bucket.id_
         )
 
+    def get_upload_url(self, bucket_id):
```
ppolewicz: this belongs in the `Bucket` class

ppolewicz:
Every `B2RawApi` method needs `api_url` (retrieved by calling `api.account_info.get_api_url()`) and an auth_token (retrieved by calling `api.account_info.get_account_auth_token()`), so the whole `Bucket` and `B2Api` classes are performing that retrieval and passing over and over.

This could be improved if `B2RawApi` constructor would require an `account_info` object. Then the methods of `B2RawApi` could call `self.account_info.get_api_url()` and `self.account_info.get_account_auth_token()`.

The change outlined in this comment can be done in a separate PR (and I will be happy to do it), but it should be done before unit tests are added (as such change would break all of them).

b2/b2.py
bwbeach: Good point.  I'll move those out of the loop.

bwbeach:
I'm OK with passing the API URL and the auth token to the B2RawApi each time.  That way the logic for calling b2_authorize_account, and for dealing with the cached account info, can be above the B2RawApi.  My goal is to have B2RawApi be a clean and very thin wrapper for the APIs that B2 provides.  That will make it easier to provide an alternate implementation for the unit tests to use.

In fact, I still plan on moving the upload and download calls into B2RawApi.

b2/b2.py
bwbeach: Yes it does.  I'll move it.

PR #82, opened 2016-03-07
Add automatic reauthorization and B2Session layer

ppolewicz:
This is not finished - many lines of code which retrieve and pass token must still be modified


bwbeach:
I like the magic.  It makes the code in B2Api a lot cleaner.

PR #83, opened 2016-03-08
Initial work on a unit-testable version of the sync command.

bwbeach:
The new code is not used yet, but I think it's a good time to merge.

I'm working towards a sync implementation that can be multi-threaded, and that has clean layering so that it's testable.  The work so far includes folder classes (local and B2) that have a common interface, and a `make_sync_folder_sync_actions` function that figures out what needs to happen to get two folders in sync.

### DIFF
--- a/b2/b2.py
+++ b/b2/b2.py
@@ -1574,13 +1574,13 @@ def download_file_by_id_helper(
         print('checksum matches')
 
 
-
 class File(object):
     """
     Holds information about one file in a folder.
 
     The name is relative to the folder in all cases.
     """
+
     def __init__(self, name, id, sha1, content_type, file_info):
         self.name = name
         self.id = id
@@ -1591,7 +1591,6 @@ class File(object):
 
 @six.add_metaclass(ABCMeta)
 class Folder(object):
-
     """
     Interface to a folder full of files, which might be a B2 bucket,
     a virtual folder in a B2 bucket, or a directory on a local file
@@ -1614,6 +1613,7 @@ class LocalFolder(Folder):
     """
     Folder interface to a directory on the local machine.
     """
+
     def __init__(self, root):
         self.root = os.path.abspath(root)
         self.relative_paths = self._get_all_relative_paths(self.root)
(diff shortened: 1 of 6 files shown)


b2/b2.py
```diff
@@ -1574,6 +1574,308 @@ def download_file_by_id_helper(
...
+    """
+    Holds information about one version of a file:
+
+       id - The B2 file id, or the local full path name
+       mod_time - modification time, in seconds
+       action - "hide" or "upload" (never "start")
+    """
+
+    def __init__(self, id, mod_time, action):
+        self.id = id
```
ppolewicz: `id` is a special word in Python. `id_` would be used by convention.

b2/b2.py
```diff
@@ -1574,6 +1574,308 @@ def download_file_by_id_helper(
...
+            current_a = next_or_none(iter_a)
+        elif current_a.name < current_b.name:
+            yield (current_a, None)
+            current_a = next_or_none(iter_a)
+        elif current_b.name < current_a.name:
+            yield (None, current_b)
+            current_b = next_or_none(iter_b)
+        else:
+            assert current_a.name == current_b.name
+            yield (current_a, current_b)
```
ppolewicz: This statement can be moved to first line of the loop and other `yield` can be deleted

test/test_sync.py
```diff
@@ -0,0 +1,164 @@
...
+
+    def __exit__(self, exc_type, exc_val, exc_tb):
+        shutil.rmtree(self.dirpath)
+        return None  # do not hide exception
+
+
+class TestLocalFolder(unittest.TestCase):
+    def test_slash_sorting(self):
+        # '/' should sort between '.' and '0'
+        with TempDir() as dir:
```
ppolewicz: `dir` is a special word in Python

test/test_sync.py
```diff
@@ -0,0 +1,164 @@
...
+            create_files(dir, ['hello.', 'hello/a', 'hello/b', 'hello0'])
+            folder = LocalFolder(dir)
+            files = list(folder.all_files())
+            names = list(map(lambda f: f.name, files))
```
ppolewicz: `names = [f.name for f in fields]`

b2/b2.py
bwbeach: right.

b2/b2.py
bwbeach: I don't think so.  In the case where `current_a` has file `a.txt` and `current_b` has `b.txt`, it needs to yield `(current_a, None)`

test/test_sync.py
bwbeach: I've been coding in Java too much.  :-(

test/test_sync.py
bwbeach: yes.

bwbeach:
Thanks for the careful review.

PR #88, opened 2016-03-12
Implement Bucket.list_unfinished_large_files()

bwbeach:


b2/b2.py
```diff
@@ -808,6 +808,34 @@ def list_file_versions(self, start_filename=None, start_file_id=None, max_entrie
...
+        account_info = self.api.account_info
+        batch_size = batch_size or 100
+        while True:
+            batch = self.api.raw_api.list_unfinished_large_files(
+                account_info.get_api_url(), account_info.get_account_auth_token(), self.id_,
+                start_file_id, batch_size
+            )
+            for file_dict in batch['files']:
+                yield UnfinishedLargeFile(file_dict)
+            start_file_id = batch['nextFileId']
```
ppolewicz: `start_file_id = batch.get('nextFileId')` so that `None` is returned when the key is missing, as expected in the following code, instead of `KeyError`

bwbeach: Yes.  That's cleaner.  The API always puts None there if it's empty, but it's better not to depend on it.

PR #89, opened 2016-03-14
Unit tests for ConsoleTool

bwbeach:


b2/b2.py
```diff
@@ -787,26 +651,19 @@ def ls(
...
-        url = url_for_api(self.api.account_info, 'b2_list_file_versions')
-        params = {
-            'bucketId': self.id_,
-            'startFileName': start_filename,
-            'startFileId': start_file_id,
-            'maxFileCount': max_entries,
-        }
-        return post_json(url, params, auth_token)
+        account_info = self.api.account_info
+        return self.api.raw_api.list_file_versions(
```
ppolewicz: Use `self.api.session.list_file_versions` and skip first two arguments

test/test_console_tool.py
```diff
@@ -0,0 +1,291 @@
...
+class TestConsoleTool(unittest.TestCase):
+    def setUp(self):
+        self.account_info = StubAccountInfo()
+        self.cache = InMemoryCache()
+        self.raw_api = RawSimulator()
+        self.b2_api = B2Api(self.account_info, self.cache, self.raw_api)
+
+    def test_authorize_and_clear_account(self):
+        # Initial condition
+        assert (self.account_info.get_account_auth_token() is None)
```
ppolewicz: Unnecessary parenthesis

ppolewicz:
The testcases should be split to small and well named ones. Each of them should contain just one `self._run_command()` call so that it is easier to quickly judge what happened by looking at the list of failures.

b2/b2.py
bwbeach: Yes!

test/test_console_tool.py
bwbeach: fixed.

bwbeach:
There is a trade-off between breaking up the tests into smaller, more targeted tests, and re-using the state for multiple tests.  I'll work on breaking these up next time I revisit them.

PR #90, opened 2016-03-15
Break up b2.py

bwbeach:


b2/account_info.py
```diff
@@ -0,0 +1,318 @@
...
+######################################################################
+
+import json
+import os
+import stat
+from abc import (ABCMeta, abstractmethod)
+
+import six
+
+from .exception import (MissingAccountData)
```
ppolewicz: unnecessary parenthesis

b2/account_info.py
```diff
@@ -0,0 +1,318 @@
...
+# Copyright 2016 Backblaze Inc. All Rights Reserved.
+#
+# License https://www.backblaze.com/using_b2_code.html
+#
+######################################################################
+
+import json
+import os
+import stat
+from abc import (ABCMeta, abstractmethod)
```
ppolewicz: unnecessary parenthesis

b2/account_info.py
```diff
@@ -0,0 +1,318 @@
...
+        flags = os.O_WRONLY | os.O_CREAT | os.O_TRUNC
+        if os.name == 'nt':
+            flags |= os.O_BINARY
+        with os.fdopen(os.open(self.filename, flags, stat.S_IRUSR | stat.S_IWUSR), 'wb') as f:
+            # is there a cleaner way to do this that works in both Python 2 and 3?
+            json_bytes = json.dumps(self.data, indent=4, sort_keys=True).encode('utf-8')
+            f.write(json_bytes)
+
+
+class StubAccountInfo(AbstractAccountInfo):
```
ppolewicz: I think this should be kept somewhere in `tests/` - we don't want to expose it to the user next to `StoredAccountInfo`

b2/api.py
```diff
@@ -0,0 +1,188 @@
...
+#
+# File: b2/api.py
+#
+# Copyright 2016 Backblaze Inc. All Rights Reserved.
+#
+# License https://www.backblaze.com/using_b2_code.html
+#
+######################################################################
+
+from .account_info import (StoredAccountInfo)
```
ppolewicz: unnecessary parenthesis here and below

b2/account_info.py
bwbeach: Yes.  They're unnecessary.  I'm undecided whether I like the consistency of always using parens on `from ... import ...` or not.  If you don't like them, I'm OK with not using them.

I was getting tired of adding and removing them as the import lists got bigger and smaller.

bwbeach: I think yapf adds parens when it goes multi-line, and doesn't remove them when i shrinks.

b2/download_dest.py
```diff
@@ -0,0 +1,124 @@
...
+import os
+from abc import (ABCMeta, abstractmethod)
+
+import six
+
+from .progress import (StreamWithProgress)
+
+
+@six.add_metaclass(ABCMeta)
+class DownloadDest(object):
```
ppolewicz: I suggest to rename it to `AbstractDownloadDestination`

b2/raw_api.py
```diff
@@ -0,0 +1,458 @@
...
+            error_dict = json.loads(e.data.decode('utf-8'))
+            raise map_error_dict_to_exception(e, error_dict, params)
+        except ValueError:
+            v = sys.exc_info()
+            error_dict = {'error_decoding_json': v}
+            raise FatalError('error decoding JSON when handling an exception', [e_backup, v],)
+
+
+@six.add_metaclass(ABCMeta)
+class RawApi(object):
```
ppolewicz: I suggest to rename to `AbstractRawApi` for consistency with other abstract classes that we have

b2/sync.py
```diff
@@ -0,0 +1,317 @@
...
+
+import os
+import time
+from abc import (ABCMeta, abstractmethod)
+
+import six
+
+
+@six.add_metaclass(ABCMeta)
+class Action(object):
```
ppolewicz: `AbstractAction`

b2/sync.py
```diff
@@ -0,0 +1,317 @@
...
+
+    def latest_version(self):
+        return self.versions[0]
+
+    def __repr__(self):
+        return 'File(%s, [%s])' % (self.name, ', '.join(map(repr, self.versions)))
+
+
+@six.add_metaclass(ABCMeta)
+class Folder(object):
```
ppolewicz: `AbstractFolder`

PR #91, opened 2016-03-16
Fix upload large files

bwbeach:
I found a couple bugs when I tried uploading to the staging servers at Backblaze.


ppolewicz:
reviewed ok

PR #93, opened 2016-03-17
Large files: commands for listing unfinished files & parts, and canceling.

bwbeach:


b2/console_tool.py
```diff
@@ -258,6 +286,24 @@ def test_raw_api(self, args):
 
     # bucket
 
+    def cancel_all_unfinished_large_files(self, args):
+        if len(args) != 1:
+            return self._usage_and_fail()
+        bucket_name = args[0]
+        bucket = self.api.get_bucket_by_name(bucket_name)
+        for file in bucket.list_unfinished_large_files():
```
ppolewicz: `file` is a reserved keyword in python. I would use `file_` instead.

b2/console_tool.py
```diff
@@ -258,6 +286,24 @@ def test_raw_api(self, args):
 
     # bucket
 
+    def cancel_all_unfinished_large_files(self, args):
```
ppolewicz: This is logic implemented in `ConsoleTool`. I think it should be done in Bucket and just called from here, like other functions (except sync)

b2/part.py
```diff
@@ -0,0 +1,41 @@
...
+        self.content_length = content_length
+        self.content_sha1 = content_sha1
+
+    def __str__(self):
+        return '<%s %s %s %s %s>' % (
+            self.__class__.__name__, self.file_id, self.part_number, self.content_length,
+            self.content_sha1
+        )
+
+    def __repr__(self):
```
ppolewicz: if __repr__ is defined, and __str__ is not, the object will behave as though __str__=__repr__. Therefore move the implementation to __repr__ and remove__str__

b2/raw_simulator.py
```diff
@@ -241,6 +278,10 @@ def list_file_versions(self, start_file_name=None, start_file_id=None, max_file_
                     break
         return dict(files=result_files, nextFileName=next_file_name, nextFileId=next_file_id)
 
+    def list_parts(self, file_id, start_part_number, max_part_count):
+        file = self.file_id_to_file[file_id]
```
ppolewicz: `file` is a reserved keyword in python

test/test_console_tool.py
```diff
@@ -110,6 +112,28 @@ def test_buckets(self):
...
+        bucket = self.b2_api.get_bucket_by_name('my-bucket')
+        file = bucket.start_large_file('file1', 'text/plain', {})
+        self._run_command(['cancel_large_file', file.file_id], '9999 canceled\n', '', 0)
+
+    def test_cancel_all_large_file(self):
+        self._authorize_account()
+        self._create_my_bucket()
+        bucket = self.b2_api.get_bucket_by_name('my-bucket')
+        bucket.start_large_file('file1', 'text/plain', {})
+        bucket.start_large_file('file1', 'text/plain', {})
```
ppolewicz: Same filename as above. Is it intentional in this test?

b2/console_tool.py
bwbeach: will fix.

b2/console_tool.py
bwbeach: I can argue that both ways.  I guess the question is whether other developers using the library will want that functionality, and the answer is probably yes.

b2/part.py
bwbeach: Cool.  I didn't know that.

test/test_console_tool.py
bwbeach: It doesn't matter for the test, but would be clearer with different file names.

PR #99, opened 2016-03-29
Interprocess locking

ppolewicz:


bwbeach:
Your code looks fine.  Looks like the build failed because `TEST_ACCOUNT_ID` was not set.

PR #102, opened 2016-03-31
Make yapf run more quietly.

bwbeach:


.travis.yml
```diff
@@ -36,5 +36,5 @@ script:
   - python setup.py install
   - nosetests -w test
   - if [ -n "$TEST_ACCOUNT_ID" ]; then python -m b2.__main__ test_raw_api; fi
-  - if [ -n "$SCA" ]; then yapf -r b2; fi
+  - if [ -n "$SCA" ]; then yapf -r b2 > /dev/null; fi
```
ppolewicz: `--diff` should be used instead of redirecting to `/dev/null` - this way we will see the violation in Travis log.

PR #103, opened 2016-03-31
Tiny wrapper for requests

bwbeach:
Not ready to merge yet, but would like comments.

### DIFF
--- a/b2/b2http.py
+++ b/b2/b2http.py
@@ -108,21 +108,6 @@ class B2Http(object):
             ...
         except B2Error as e:
             ...
-
-
-    get_content() returns a context manager object that returns
-    an object that acts like requests.Response.
-
-        try:
-            with b2_http.get_content(url, headers) as response:
-                for byte_data in response.iter_content(chunk_size=1024):
-                    ...
-        except B2Error as e:
-            ...
-
-    The response object is only guarantee to have:
-        - headers
-        - iter_content()
     """
 
     def __init__(self, requests_module=None):
@@ -134,10 +119,18 @@ class B2Http(object):
 
     def post_content_return_json(self, url, headers, data):
         """
+        Use like this:
+
+            try:
+                response_dict = b2_http.post_content_return_json(url, headers, data)
+                ...
+            except B2Error as e:
+                ...
+
         :param url: URL to call
         :param headers: Headers to send.
-        :param data: bytes (Python 3) or str (Python 2) to send
-        :return:
+        :param data: bytes (Python 3) or str (Python 2), or a file-like object, to send
+        :return: a dict that is the decoded JSON
         """
         headers['User-Agent'] = USER_AGENT
         response = _translate_errors(
@@ -149,22 +142,47 @@ class B2Http(object):
             )
         )
         try:
+            # We only get here if the status is 200 OK, and it
+            # that case the body is *always* JSON.
             return json.loads(response.content.decode('utf-8'))
         finally:
             response.close()
 
     def post_json_return_json(self, url, headers, params):
         """
+        Use like this:
+
+            try:
+                response_dict = b2_http.post_json_return_json(url, headers, params)
+                ...
+            except B2Error as e:
+                ...
+
         :param url: URL to call
         :param headers: Headers to send.
         :param params: A dict that will be converted to JSON
-        :return:
+        :return: a dict that is the decoded JSON
         """
         data = six.b(json.dumps(params))
         return self.post_content_return_json(url, headers, data)
 
     def get_content(self, url, headers):
         """
+        Fetches content from a URL.
+
+        Use like this:
+
+            try:
+                with b2_http.get_content(url, headers) as response:
+                    for byte_data in response.iter_content(chunk_size=1024):
+                        ...
+            except B2Error as e:
+                ...
+
+        The response object is only guarantee to have:
+            - headers
+            - iter_content()
+
         :param url: URL to call
         :param headers: Headers to send
         :return: Context manager that returns an object that supports iter_content()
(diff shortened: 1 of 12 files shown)


b2/b2http.py
```diff
@@ -0,0 +1,178 @@
...
+
+    except requests.ConnectionError as e0:
+        e1 = e0.args[0]
+        if isinstance(e1, requests.packages.urllib3.exceptions.MaxRetryError):
+            msg = e1.args[0]
+            if 'nodename nor servname provided, or not known' in msg:
+                # Unknown host, or DNS failing.  In the context of calling
+                # B2, this means that something is down between here and
+                # Backblaze, so we treat it like 503 Service Unavailable.
+                raise WebApiError(503, 'unknown_host', 'unable to locate host')
```
ppolewicz: This is misleading: 503 means good connection to load balancer/front server,  but backend server error/all backends broken or busy

b2/b2http.py
```diff
@@ -0,0 +1,178 @@
...
+        elif isinstance(e1, requests.packages.urllib3.exceptions.ProtocolError):
+            e2 = e1.args[1]
+            if isinstance(e2, socket.error):
+                if e2.args[1] == 'Broken pipe':
+                    # Broken pipes are usually caused by the service rejecting
+                    # an upload request for cause, so we use a 400 Bad Request
```
ppolewicz: Comment inconsistent with the code

b2/b2http.py
```diff
@@ -0,0 +1,178 @@
...
+                    # code.
+                    raise WebApiError(503, 'broken_pipe', 'unable to send entire request')
+        raise WebApiError(503, 'connection_error', 'unable to connect')
+
+    except Exception as e:
+        # Don't expect this to happen.  To get lots of info for
+        # debugging, call print_exception.
+        # print_exception(e)
```
ppolewicz: Commented code

b2/b2http.py
```diff
@@ -0,0 +1,178 @@
...
+def post(url, headers, data):
+    """
+    Posts to the given URL, translating any errors into a raised WebApiError.
+
+    :param url: The URL to post to.
+    :param headers: Headers to send.
+    :param data: Data to send: should be bytes.
+    :return: A requests object with the response.
+    """
+    return _translate_errors(lambda: requests.post(url, headers=headers, data=data))
```
ppolewicz: `functools.partial` instead of lambda

b2/b2http.py
```diff
@@ -0,0 +1,178 @@
...
+    # Generic connection error
+    print('TEST: generic connection error')
+    try:
+        post('https://api.backblaze.com:6666/bad_port', {}, six.b(chr(0)) * 10000000)
+        assert False, 'should have failed with broken pipe'
+    except WebApiError as e:
+        assert e == WebApiError(503, 'connection_error', 'unable to connect')
+
+
+_test()
```
ppolewicz: Should not be called here

b2/b2http.py
```diff
@@ -0,0 +1,178 @@
...
+        super(WebApiError, self).__init__(status, code, message)
+        self.status = status
+        self.code = code
+        self.message = message
+
+    def __repr__(self):
+        return "WebApiError(%d, '%s', '%s')" % (self.status, self.code, self.message)
+
+    def __eq__(self, other):
+        return repr(self) == repr(other)
```
ppolewicz: This seems weird, exceptions don't usually have this. It is needed for tests, but checking `code` field is enough there. In test we want to check behavior, not identity.

bwbeach: I'll go with `code`.

b2/b2http.py
bwbeach: Why?  I find the lambda here concise and clear.

b2/b2http.py
bwbeach: 503 can also be returned by a proxy in between you and the server.  If you think of this code as a proxy, it makes sense.

b2/b2http.py
bwbeach: Found a [StackOverflow article](https://stackoverflow.com/questions/3252228/python-why-is-functools-partial-necessary).  I'm willing to follow Guido on this one.

(And yes, I did do a lot of Lisp programming early in my career.)

b2/b2http.py
```diff
@@ -0,0 +1,230 @@
...
+    The operations supported are:
+       - post_json_return_json
+       - post_content_return_json
+       - get_content
+
+    The methods that return JSON either return a Python dict  or
+    raise a subclass of B2Error.  They can be used like this:
+
+        try:
+            response_dict = b2_http.post_json_return_json(url, headers, params)
```
ppolewicz: This docstring should be for method, not for the class

b2/b2http.py
```diff
@@ -0,0 +1,230 @@
...
+        response = _translate_errors(
+            functools.partial(
+                requests.post,
+                url,
+                headers=headers,
+                data=data
+            )
+        )
+        try:
+            return json.loads(response.content.decode('utf-8'))
```
ppolewicz: Is empty or non-json response handled?

b2/raw_api.py
```diff
@@ -223,7 +123,8 @@ def _post_json(self, base_url, api_name, auth, **params):
         :return:
         """
         url = base_url + '/b2api/v1/' + api_name
-        return post_json(url, params, auth)
+        headers = dict(Authorization=auth)
```
ppolewicz: We usually use `{'Authorization': auth}` syntax

b2/b2http.py
bwbeach: OK.  It will duplicate some information, but you're right.

b2/b2http.py
bwbeach: No.  I'll add a comment.

PR #105, opened 2016-03-31
Simple thread pool class.

bwbeach:
I didn't find an obvious library to use, so I wrote a super-simple thread pool.

### DIFF
--- a/b2/thread_pool.py
+++ b/b2/thread_pool.py
@@ -17,10 +17,12 @@ from six.moves.queue import Queue
 # this is added to the task queue to tell the threads to stop
 SHUT_DOWN_TOKEN = 'shut_down'
 
+
 class ThreadInThreadPool(threading.Thread):
     """
     Runs tasks from the queue until getting a SHUT_DOWN_TOKEN.
     """
+
     def __init__(self, pool, queue):
         super(ThreadInThreadPool, self).__init__()
         self.pool = pool
@@ -108,7 +110,3 @@ class ThreadPool(object):
         thread = ThreadInThreadPool(self, self.queue)
         thread.start()
         return thread
-
-
-
-
--- a/b2/thread_pool.py
+++ b/b2/thread_pool.py
@@ -1,0 +1,114 @@
+######################################################################
+#
+# File: b2/thread_pool.py
+#
+# Copyright 2016 Backblaze Inc. All Rights Reserved.
+#
+# License https://www.backblaze.com/using_b2_code.html
+#
+######################################################################
+
+from __future__ import print_function
+
+import threading
+from six.moves import range
+from six.moves.queue import Queue
+
+# this is added to the task queue to tell the threads to stop
+SHUT_DOWN_TOKEN = 'shut_down'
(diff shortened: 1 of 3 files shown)


b2/thread_pool.py
```diff
@@ -0,0 +1,112 @@
...
+######################################################################
+
+from __future__ import print_function
+
+import threading
+from six.moves import range
+from six.moves.queue import Queue
+
+# this is added to the task queue to tell the threads to stop
+SHUT_DOWN_TOKEN = 'shut_down'
```
ppolewicz: I suggest to change it to None or an object - comparing strings via `is` operator is not recommended

b2/thread_pool.py
```diff
@@ -0,0 +1,112 @@
...
+        """
+        self.queue.put(task)
+
+    def join_all(self):
+        """
+        Waits until all tasks are completed and then shuts down the threads.
+        """
+        # Tell each of the threads to stop.  Each thread will consume
+        # one of the tokens and then exit.
+        for i in range(len(self.threads)):
```
ppolewicz: Please add a comment here explaining why it is not `for t in self.threads`

b2/thread_pool.py
```diff
@@ -0,0 +1,112 @@
...
+            self.queue.put(SHUT_DOWN_TOKEN)
+        self.queue = None
```
ppolewicz: Why?

b2/thread_pool.py
```diff
@@ -0,0 +1,112 @@
...
+
+        # Wait for the threads to be done.
+        for thread in self.threads:
+            thread.join()
+        self.threads = None
```
ppolewicz: Why?

b2/thread_pool.py
```diff
@@ -0,0 +1,112 @@
...
+        pool.join_all()
+        errors = pool.get_exceptions()
+    """
+
+    def __init__(self, thread_count=10, queue_capacity=1000):
+        """
+        Initializes a new thread pool, starts the threads running,
+        and gets ready to accept tasks.
+        """
+        self.queue = Queue(maxsize=queue_capacity)
```
ppolewicz: Those fields are not a part of class interface and should have an underscore prefix

b2/thread_pool.py
bwbeach: Will do.

b2/thread_pool.py
bwbeach: `for _ in self.threads`

b2/thread_pool.py
bwbeach: So that nobody adds any tasks after calling `join_all`.  Will comment.

b2/thread_pool.py
bwbeach: Because there aren't any threads any more.  ThreadPool is intended to a use-once class.

b2/thread_pool.py
bwbeach: yes.

PR #106, opened 2016-04-01
Rethink ThreadPool class.

bwbeach:
I think that we'll need to have tasks be able to add tasks
to the queue.  This means that using the shutdown tokens
in join() won't work, because other tasks might get added
behind them.

So the join() method now uses Queue.join().  And there is a
separate shut_down() method that stops the threads after
all of the tasks have been completed.

### DIFF
--- a/b2/thread_pool.py
+++ b/b2/thread_pool.py
@@ -32,11 +32,13 @@ class ThreadInThreadPool(threading.Thread):
         while True:
             task = self.queue.get()
             if task is SHUT_DOWN_TOKEN:
+                self.queue.task_done()
                 return
             try:
                 task.run()
             except Exception as e:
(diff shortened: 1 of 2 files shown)


b2/thread_pool.py
```diff
@@ -100,6 +103,21 @@ def get_exceptions(self):
         with self._lock:
             return self._exceptions
 
+    def shut_down(self):
+        # Nobody can add any tasks any more.
+        with self._lock:
+            queue = self._queue
+            self._queue = None
```
ppolewicz: if `shut_down` will race with `add_task` or `join`, 'None has no attribute 'join' will be the error we get. I suggest using a dummy queue object instead to raise an exception ("unable to add tasks to a queue during shutdown" or something)

PR #108, opened 2016-04-04
Teach b2http to retry on server errors

bwbeach:


b2/b2http.py
```diff
@@ -117,7 +138,7 @@ def __init__(self, requests_module=None):
         """
         self.requests = requests_module or requests
 
-    def post_content_return_json(self, url, headers, data):
+    def post_content_return_json(self, url, headers, data, try_count=1):
```
ppolewicz: why 1 here?

PR #110, opened 2016-04-06
Locking and upload URL management

bwbeach:
Only one thread at a time will use an upload URL.

### DIFF
--- a/b2/account_info.py
+++ b/b2/account_info.py
@@ -85,8 +85,19 @@ class AbstractAccountInfo(object):
         pass
 
     @abstractmethod
-    def set_bucket_upload_data(self, bucket_id, upload_url, upload_auth_token):
-        pass
+    def take_bucket_upload_url(self, bucket_id):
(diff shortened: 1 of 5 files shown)


ppolewicz:
It is good, but a context manager would be a pythonic way of doing what `self._update_data(update_fcn)` is currently doing.
It should be like this:
```python
with self._update_data() as data:
```
And then locking and updating would be done internally

bwbeach:
I went back and forth on whether to use a context manager or a functional approach.  The functional approach had cleaner, more concise code.  

I did try writing the context manager code, and it's ugly.  The __enter__ method has explicitly call the __enter__ method on the Lock().  I couldn't think of a way to have the context manager use a with statement on the Lock, so by having `update_data` be a context manager, I lost the ability to use the with statement on the Lock.

ppolewicz:
I think calling `__enter__` explicitly is ok in such case.

PR #122, opened 2016-04-18
Sync: checkpoint work

bwbeach:
Sync is not done, but syncing up to B2 mostly works.  And there have been a lot of changes (almost all in sync.py), so it's a good time to merge in.

### DIFF
--- a/b2/sync.py
+++ b/b2/sync.py
@@ -466,10 +466,11 @@ def make_file_sync_actions(sync_type, source_file, dest_file, source_folder, des
         else:
             yield B2DownloadAction(source_file.name, source_file.latest_version().id_)
     if source_mod_time == 0 and dest_mod_time != 0:
-        if sync_type == 'local-to-b2':
-            yield B2DeleteAction(dest_file.name, dest_file.latest_version().id_)
-        else:
-            yield LocalDeleteAction(dest_file.latest_version().id_)
+        if args.delete:
+            if sync_type == 'local-to-b2':
+                yield B2DeleteAction(dest_file.name, dest_file.latest_version().id_)
+            else:
+                yield LocalDeleteAction(dest_file.latest_version().id_)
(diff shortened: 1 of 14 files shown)


b2/sync.py
```diff
@@ -421,39 +649,115 @@ def zip_folders(folder_a, folder_b):
...
+
+    # Case 2: Both exist and source is older
+    elif source_mod_time != 0 and source_mod_time < dest_mod_time:
+        if args.replaceNewer:
+            yield make_transfer_action(sync_type, source_file, source_folder, dest_folder)
+            transferred = True
+        elif args.skipNewer:
+            pass
+        else:
+            raise DestFileNewer('destination file is newer: %s' % (dest_file.name,))
```
ppolewicz: by our convention this message string should be in `DestFileNewer` exception class and just the filename should be passed in

test/test_sync.py
```diff
@@ -91,65 +128,281 @@ def test_two(self):
...
-                "b2_download(a.txt, id_a_100)", "local_delete(/dir/b.txt)",
-                "b2_download(e.txt, id_e_300)"
-            ], list(map(str, actions))
+class TestMakeSyncActions(unittest.TestCase):
+    def setUp(self):
+        self.reporter = MagicMock()
+
+    def test_illegal_b2_to_b2(self):
+        b2_folder = FakeFolder('b2', [])
+        try:
```
ppolewicz: this is very much like `unittest.TestCase.assertRaises`. Can we use it here?

bwbeach: assertRaises is not in 2.6.  :-(

ppolewicz: then we should implement the wrapper by ourselves in one place instead of writing it in each test

b2/sync.py
bwbeach: done.

PR #123, opened 2016-04-19
Fixed multiple InvalidAuthToken errors

svonohr:
Fixed some of the errors discussed here #7 

I also changed the InvalidAuthToken exception so that the client retries uploading after such an error. Without this I was still getting errors if I upload very small files and the auth token was invalid.

### DIFF
--- a/b2/exception.py
+++ b/b2/exception.py
@@ -118,6 +118,9 @@ class InvalidAuthToken(B2Error):
     def __str__(self):
         return 'Invalid authorization token. Server said: %s (%s)' % (self.message, self._type)
 
+    def should_retry_upload(self):
+        return True
+
 
 class MaxFileSizeExceeded(B2Error):
     def __init__(self, size, max_allowed_size):
--- a/b2/bucket.py
+++ b/b2/bucket.py
@@ -224,12 +224,10 @@ class Bucket(object):
         unfinished large file in the bucket, starting at the
         given file.
         """
-        account_info = self.api.account_info
         batch_size = batch_size or 100
         while True:
-            batch = self.api.raw_api.list_unfinished_large_files(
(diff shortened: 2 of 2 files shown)


b2/exception.py
```diff
@@ -118,6 +118,9 @@ def __init__(self, message, _type):
     def __str__(self):
         return 'Invalid authorization token. Server said: %s (%s)' % (self.message, self._type)
 
+    def should_retry_upload(self):
```
ppolewicz: I think `should_retry_http` should also return `True` here (right?)

ppolewicz:
`self.api.raw_api` is used in 3 more places. Will you fix those too with further commits?

svonohr:
Those places which still use `self.api.raw_api` need to stay this way because they use a different request URL. They also have additional logic which will retry the request (except for `download_file_by_name`).

The `should_retry_http` property should not be enable for invalid auth exceptions since this will just retry with the same auth key again, which is unlikely to succeed.

ppolewicz:
See #124 - I think I managed to reauthorize `download_file_by_name` too.

ppolewicz:
Thank you for your help @svonohr!

By the way, maybe you'd like to contribute something more from time to time? You mentioned that you don't feel like python expert and we happen to have seasoned developers who would be happy to offer friendly advice to a contributor. We can make it in small tasks such as the one you have just tackled.

svonohr:
I'm having a lot of stuff to do with my own projects, but maybe I can do a few small things. One larger feature that I am personaly interested in is encrypting files on the client side during upload. At the moment I'm using GPG to encrypt files, but this requires creating a possibly large temporary file.

bwbeach:
Encryption would be a great new feature.

ppolewicz:
Doesn't seem like a lot of code too. Would you create a new Issue for it so that we can discuss the design?

PR #124, opened 2016-04-19
Finish reauthorization

ppolewicz:
This is built on top of #123 which was contributed by @svonohr


ppolewicz:
Currently it includes all commits from #123, so both can be closed upon merging this one

svonohr:
Looks good to me. I've verified that `download_file_by_name` was indeed suffering from missing reauthentication and that it was fixed by your changes.

PR #130, opened 2016-04-21
Use sqlite to store account info in .b2_account_info

### LINKED ISSUE #128: dependency problem on Windows: win32con
On windows, the `portalocker` package has a dependency on `win32con`, which is not available through pip.  It has its own installer.

I don't want to force people to run yet another windows installer to get `win32con`.

I tried installing a fresh python from python.org, and then installing `b2` from pypi.  When I try to run b2, I get this error:

    ImportError: No module named 'win32con'

bwbeach:
Fix #128

### DIFF
--- a/b2/account_info.py
+++ b/b2/account_info.py
@@ -11,8 +11,7 @@
 import collections
 import json
 import os
-import portalocker
-import stat
+import sqlite3
 import threading
 from abc import (ABCMeta, abstractmethod)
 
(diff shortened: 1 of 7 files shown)


b2/account_info.py
```diff
@@ -114,123 +141,258 @@ def clear_large_file_upload_urls(self, file_id):
...
+                        insert_statement = """
+                            INSERT INTO account
+                            (account_id, application_key, account_auth_token, api_url, download_url, minimum_part_size, realm)
+                            values (?, ?, ?, ?, ?, ?, ?);
+                        """
+
+                        conn.execute(insert_statement, tuple(data[k] for k in keys))
+                    # all is happy now
+                    return
+        except:
```
ppolewicz: this will catch CTRL+C if the hard drive is slow/overloaded during f.read(). User gets impatient, wants to interrupt to check what his system is doing using `top` and the code below deletes his account settings.

I suggest that we enumerate the exceptions we really want to catch here or stop automatic recovery, only advising the user that we could not open it and he might need to remove the configuration file himself if it is corrupted.

b2/account_info.py
```diff
@@ -114,123 +141,258 @@ def clear_large_file_upload_urls(self, file_id):
...
+            pass
+
+        # Remove the corrupted file and create a new database
```
ppolewicz: if this block will stay, it should go to the `except` clause and `else` (under `with`) to prevent fall-through due to future changes in the code above, from deleting a file.

b2/account_info.py
bwbeach: Good ideas.  I'll change it to report a corrupted file, but not delete it.

b2/account_info.py
```diff
@@ -114,123 +141,257 @@ def clear_large_file_upload_urls(self, file_id):
...
 
-        self._update_data(update_fcn)
+    def _get_account_info_or_raise(self, column_name):
+        try:
+            with self._get_connection() as conn:
+                cursor = conn.execute('SELECT %s FROM account;' % (column_name,))
+                value = cursor.fetchone()[0]
+                return value
+        except Exception as e:
+            raise MissingAccountData(str(e))
```
ppolewicz: Should be `column_name` instead of `str(e)` to keep th let interface

PR #134, opened 2016-04-27
Encrypting files for uploads

svonohr:
I've added some encryption classes and functions. The `UploadEncryptedSourceWrapper` class can wrap any of the existing upload sources and encrypt them while reading/uploading. To test this you can use:
`upload_source = UploadEncryptedSourceWrapper(upload_source, EncryptionContext())`
At the moment just a dummy key is used for encryption. Later the encryption context should be managed somewhere in the bucket. I'm not exactly sure where to put this. How to we initialize an encrypted bucket or how do hide the fact that the bucket is actually encrypted when using the API?

### DIFF
--- a/b2/encryption.py
+++ b/b2/encryption.py
@@ -139,7 +139,7 @@ class EncryptionContext(object):
     def block_count(self, file_size):
         return (file_size + self.block_size - 1) // self.block_size  # divide rounding up
 
-    def encrypt_filename(filename, folder=''):
+    def encrypt_filename(self, filename, folder=''):
         # derive multiple keys
         enc_key = derive_key(self.master_key, b'filename', 1)
         hmac_key = derive_key(self.master_key, b'filename_hmac', 1)
@@ -158,7 +158,7 @@ class EncryptionContext(object):
         filenamePadded = padder.update(filename) + padder.finalize()
         return iv + encryptor.update(filenamePadded) + encryptor.finalize()
 
-    def decrypt_filename(masterKey, filename):
+    def decrypt_filename(self, filename):
         # get key and iv
         enc_key = derive_key(self.master_key, b'filename', 1)
         iv = filename[0:16]
--- a/b2/encryption.py
+++ b/b2/encryption.py
@@ -1,0 +1,174 @@
+######################################################################
+#
+# File: b2/encryption.py
+#
+# Copyright 2016 Backblaze Inc. All Rights Reserved.
+#
+# License https://www.backblaze.com/using_b2_code.html
+#
+######################################################################
+
+import os
+import struct
+
+from cryptography.hazmat.backends import default_backend
+from cryptography.hazmat.primitives import hashes, hmac, padding
+from cryptography.hazmat.primitives.kdf.pbkdf2 import PBKDF2HMAC
+from cryptography.hazmat.primitives.ciphers import (Cipher, algorithms, modes)
+
+
+def derive_key(passphrase, salt, iterations=500000):
+    kdf = PBKDF2HMAC(
+        algorithm=hashes.SHA256(),
+        length=16,
+        salt=salt,
+        iterations=iterations,
+        backend=default_backend()
+    )
+    return kdf.derive(passphrase)
+
+
+def pack_iv(iv):
+    return struct.pack('>III', (iv >> 64) & 0xFFFFFFFF, (iv >> 32) & 0xFFFFFFFF, iv & 0xFFFFFFFF)
+
+
+def unpack_iv(piv):
+    iv = struct.unpack('>III', piv)
+    return (iv[0] << 64) | (iv[1] << 32) | iv[2]
+
+
+class EncryptingFileStream(object):
+    def __init__(self, upload_source, crypto_file):
+        self.stream = upload_source.open()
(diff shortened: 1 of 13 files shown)


svonohr:
I don't know how to fix the build errors for pypy.

ppolewicz:
Let me look into it

ppolewicz:
Try rebasing to `pypy4` branch

svonohr:
I'm not sure if this does the correct thing. At least the checks are sucessful.

ppolewicz:
It tests pypy2 twice. I have updated my branch so that it upgrades pypy3 properly. Please retarget again.

ppolewicz:
Ouch. Can you try if it works with Cryptography 0.9.3?

svonohr:
At least the pypy builds succeed now. I think cryptography doesn't support pypy3 properly, since it requests a version which does not exist.

ppolewicz:
Cryptography 0.9.3 supports pypy and pypy3, but does not support python 3.6, while 1.0.0 supports new pypy2, python 3.6 and does not support pypy3.
I think we should discuss this with Cryptography maintainers. If they would add support for pypy3 in their 1.0 branch or would release 0.9.4 with support for python 3.6 (current head), we'd be all set. Will you try to talk to them, or shall I?

svonohr:
Maybe you can talk to them? I'm not exactly sure what we need.

svonohr:
Should we ignore or disable the pypy3 test for now, so that we can continue working on the encryption? I was thinking that we could have an `EncryptedBucket` class, which inherits from the original `Bucket` class. The overridden functions then handles the encryption/decryption when passing in arguments or receiving return values. Or do you have a different idea?

ppolewicz:
@bwbeach we need to temporarily drop support for pypy3 or python nightly (3.6). Which one should we drop?

bwbeach:
I don't think I care very much.  3.6 is not scheduled for release until the end of 2016.  I don't know how many people use pypy3, but my impression is that it's not done yet.

My inclination is to keep the main branch working, so I would say to keep 3.6 working.

ppolewicz:
@svonohr do you know how to proceed?

ppolewicz:
I have created a ticket in Cryptography github: https://github.com/pyca/cryptography/issues/2880

svonohr:
I'll just disable pypy3 testing for now, as bwbeach suggested. I've asked if it was a good idea to have an `EncryptedBucket` class. Should I start implementing this or do you have a better approach?

bwbeach:
Sorry I haven't been keeping up.  It's been a week with lots of meetings.  :-(

Do we have a proposal for what the command-line options will look like, or is the proposal so far at the library level?

svonohr:
So far I'm only concerned with the library level. For the command-line options I have no good idea. It should be something that is not too intrusive, i.e., you shouldn’t have to specify that you want encryption for every command.

ppolewicz:
Encryption in current implementation will be a bucket-level setting. One could specify that the bucket should be encrypted during bucket creation with `--encryptionkey "abc"`.

Things to consider *for the future* (separate issues):
 1. Changing bucket encryption status from non-encrypted to encrypted
 1. Changing bucket encryption status from encrypted to non-encrypted

I am worried about multi-client environment where one endpoint encrypts a bucket, but another one doesn't know about it. How do we enforce all endpoints to use encryption? Maybe we can invalidate all tokens and make endpoints check for presence of encryption marker during requisition of a new token?

bwbeach:
I think that each file should have the information about how it was encrypted, and the name of the key used to encrypt it.  

When you create an EncryptedBucket object in Python, you can provide a set of keys (each with a name) for decrypting, and a key to use for encrypting files that are uploaded.

ppolewicz:
if they are identified by names, lets make it a dictionary of keys instead of a set.

ppolewicz:
by the way, cryptography maintainers already fixed their build on Python 3.6 a while ago, but fix was not released to pypi yet.

#142 shall integrate the necessary environment changes into master, so that you can safely rebase and continue development of encryption support

ppolewicz:
Cryptography developers say they released the fix with 1.1 release.

@svonohr, could you please retarget on cryptography 1.1 and see if everything works properly there?

b2/encryption.py
```diff
@@ -0,0 +1,174 @@
...
+from cryptography.hazmat.primitives.ciphers import (Cipher, algorithms, modes)
+
+
+def derive_key(passphrase, salt, iterations=500000):
+    kdf = PBKDF2HMAC(
+        algorithm=hashes.SHA256(),
+        length=16,
+        salt=salt,
+        iterations=iterations,
+        backend=default_backend()
```
bwbeach: Maybe this should specify a backend, in case the library changes its mind?

The doc says:

    The default backend will change over time as we implement new backends and the libraries we use in those backends changes.

ppolewicz:
Tests pass now.

b2/encryption.py
svonohr: I think they won't change the backend to something which supports less features. The only really useful backend at the moment seems to be OpenSSL. But I can change the code so that we can easily change the backend when required.

ppolewicz:
This PR is replaced by #146

Issue #137, opened 2016-04-28
Resuming large file uploads


bwbeach:
When uploading a large file, it's annoying to have it fail after uploading most of the parts, and then have to start the whole process over again.  It would be better to check and see if there is a matching upload (file name, content type, etc.) already in progress with parts that match the file being uploaded, and then continuing from there.

svonohr:
This is a very useful feature. I will try to implement it. How should multiple unfinished files with the same name be handled? Do we just check the filename, file info and part size or also verify the SHA-1 hash of already uploaded parts?

bwbeach:
I think it should verify the SHA1 of the uploaded parts.  If they don't match, you'll get an error at the end, and it will be frustrating.

bwbeach:
It could be that two invocations try to resume the file.  I'm not sure how to prevent that, or how much to worry about it.

PR #139, opened 2016-04-28
Resuming large file uploads

svonohr:
Implemented resuming large file uploads as discussed in #137 
At the moment SHA1 hashes might be calculated more than once for individual parts, but I didn't want to change the structure of the classes to much. I've not done extensive testing, since my upload bandwidth doesn't allow this at the moment. I will test this some more tomorrow.

### DIFF
--- a/b2/bucket.py
+++ b/b2/bucket.py
@@ -386,7 +386,8 @@ class Bucket(object):
                 part_index + 1,  # part number
                 part_range,
                 upload_source,
-                large_file_upload_state
+                large_file_upload_state,
+                finished_parts
             ) for (part_index, part_range) in enumerate(part_ranges)
         ]
 
@@ -428,11 +429,26 @@ class Bucket(object):
 
                 # Return first matched file
                 return (file, finished_parts)
-        return (None, None)
+        return (None, [])
 
     def _upload_part(
-        self, file_id, part_number, part_range, upload_source, large_file_upload_state
+        self,
+        file_id,
+        part_number,
+        part_range,
+        upload_source,
+        large_file_upload_state,
+        finished_parts=[]
     ):
+        # Check if this part was uploaded before
+        for part in finished_parts:
+            if part.part_number == part_number:
+                # Report this part finished
+                large_file_upload_state.update_part_bytes(part.content_length)
+
+                # Return SHA1 hash
+                return {'contentSha1': part.content_sha1}
+
         # Compute the SHA1 of the part
         (offset, content_length) = part_range
         with upload_source.open() as f:
--- a/b2/bucket.py
+++ b/b2/bucket.py
@@ -368,8 +368,14 @@ class Bucket(object):
         # Select the part boundaries
         part_ranges = choose_part_ranges(content_length, minimum_part_size)
 
-        # Tell B2 we're going to upload a file
-        unfinished_file = self.start_large_file(file_name, content_type, file_info)
+        # Check for unfinished files with same name
+        (unfinished_file, finished_parts) = self._find_unfinshed_file(
+            upload_source, file_name, file_info, part_ranges
+        )
+
+        # Tell B2 we're going to upload a file if necessary
+        if unfinished_file is None:
+            unfinished_file = self.start_large_file(file_name, content_type, file_info)
         file_id = unfinished_file.file_id
 
         # Tell the executor to upload each of the parts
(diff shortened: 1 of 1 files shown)


b2/bucket.py
```diff
@@ -393,6 +399,37 @@ def _upload_large_file(
         response = self.api.session.finish_large_file(file_id, part_sha1_array)
         return FileVersionInfoFactory.from_api_response(response)
 
+    def _find_unfinshed_file(self, upload_source, file_name, file_info, part_ranges):
+        for file in self.list_unfinished_large_files():
+            if file.file_name == file_name and file.file_info == file_info:
+                files_match = True
+                finished_parts = []
```
bwbeach: I think it would make sense for finished_parts to be a dict that maps from part number to the part object.  That will make the lookup quicker during upload.

b2/bucket.py
```diff
@@ -393,6 +399,37 @@ def _upload_large_file(
         response = self.api.session.finish_large_file(file_id, part_sha1_array)
         return FileVersionInfoFactory.from_api_response(response)
 
+    def _find_unfinshed_file(self, upload_source, file_name, file_info, part_ranges):
```
bwbeach: A doc string on this method would be nice.

bwbeach:
Nice work!

I'd like to code up a unit test for this.  What's the right way to handle that in Github.  Should I make a branch in this repository with your changes in it?

svonohr:
A branch for this would be fine.

bwbeach:
I made a branch.  I'll add unit tests either today or tomorrow morning.

bwbeach:
I added @svonohr as a collaborator, so you can push directly to branches in this repository.  Please do not push directly to master.

b2/bucket.py
```diff
@@ -393,9 +400,55 @@ def _upload_large_file(
...
+        for file in self.list_unfinished_large_files():
```
ppolewicz: `file` is a reserved keyword in python. Please use `file_` or some other name instead.

b2/bucket.py
```diff
@@ -368,8 +368,14 @@ def _upload_large_file(
         # Select the part boundaries
         part_ranges = choose_part_ranges(content_length, minimum_part_size)
 
-        # Tell B2 we're going to upload a file
-        unfinished_file = self.start_large_file(file_name, content_type, file_info)
+        # Check for unfinished files with same name
+        (unfinished_file, finished_parts) = self._find_unfinshed_file(
```
ppolewicz: unnecessary parenthesis (they make yapf slower)

b2/bucket.py
```diff
@@ -393,9 +400,55 @@ def _upload_large_file(
...
         return FileVersionInfoFactory.from_api_response(response)
 
+    def _find_unfinshed_file(self, upload_source, file_name, file_info, part_ranges):
+        for file in self.list_unfinished_large_files():
+            if file.file_name == file_name and file.file_info == file_info:
+                files_match = True
+                finished_parts = []
+                for part in self.list_parts(file.file_id):
+                    # Compare part sizes
+                    (offset, part_length) = part_ranges[part.part_number - 1]
```
ppolewicz: unnecessary parenthesis (they make yapf slower and the code less readable (for me))

b2/bucket.py
```diff
@@ -393,9 +400,60 @@ def _upload_large_file(
         response = self.api.session.finish_large_file(file_id, part_sha1_array)
         return FileVersionInfoFactory.from_api_response(response)
 
+    def _find_unfinished_file(self, upload_source, file_name, file_info, part_ranges):
+        """
+        Find an unfinished file which may be used to resume a large file upload. The
+        file is found using the filename and comparing the uploaded parts against
+        the local file.
+        """
+        for file in self.list_unfinished_large_files():
```
ppolewicz: `file` is a reserved keyword in python. Please use `file_` or some other name instead.

b2/bucket.py
```diff
@@ -393,9 +400,60 @@ def _upload_large_file(
...
+            if file.file_name == file_name and file.file_info == file_info:
+                files_match = True
+                finished_parts = {}
+                for part in self.list_parts(file.file_id):
+                    # Compare part sizes
+                    (offset, part_length) = part_ranges[part.part_number - 1]
```
ppolewicz: unnecessary parenthesis (they make yapf slower and the code less readable (for me))

b2/bucket.py
```diff
@@ -368,8 +368,14 @@ def _upload_large_file(
         # Select the part boundaries
         part_ranges = choose_part_ranges(content_length, minimum_part_size)
 
-        # Tell B2 we're going to upload a file
-        unfinished_file = self.start_large_file(file_name, content_type, file_info)
+        # Check for unfinished files with same name
+        (unfinished_file, finished_parts) = self._find_unfinished_file(
```
ppolewicz: unnecessary parenthesis (they make yapf slower and the code less readable (for me))

b2/bucket.py
```diff
@@ -393,9 +400,60 @@ def _upload_large_file(
...
+                        sha1_sum = hex_sha1_of_stream(f, part_length)
+                    if sha1_sum != part.content_sha1:
+                        files_match = False
+                        break
+
+                    # Save part
+                    finished_parts[part.part_number] = part
+
+                # Skip not matching files or unfinished files with no uploaded parts
+                if files_match == False or len(finished_parts) == 0:
```
ppolewicz:
```python
if not files_match or not finished_parts:
```

b2/bucket.py
```diff
@@ -393,9 +400,60 @@ def _upload_large_file(
...
+                    continue
+
+                # Return first matched file
+                return (file, finished_parts)
```
ppolewicz: unnecessary parenthesis (they make yapf slower and the code less readable (for me)). Same for one line below.

b2/bucket.py
bwbeach: I like these parentheses.  I'm in the habit of always putting parens around tuples.  I find it more readable.   I guess we disagree.  :-)

b2/bucket.py
```diff
@@ -393,9 +400,60 @@ def _upload_large_file(
...
+
     def _upload_part(
-        self, file_id, part_number, part_range, upload_source, large_file_upload_state
+        self,
+        file_id,
+        part_number,
+        part_range,
+        upload_source,
+        large_file_upload_state,
+        finished_parts={}
```
ppolewicz: this dictionary is global. Common python bug.
To fix it, use `finished_parts=None` and
```python
if finished_parts is None:
    finished_parts = {}
```

svonohr:
I will fix the issues soon. How should I handle the parenthesis, since you two disagree?

ppolewicz:
However you prefer yourself. You are the tie-breaking contributor now :smile:.
I tried to find how python standard library does it and I found [this](https://github.com/python/cpython/blob/master/Lib/argparse.py#L1843). Another argument against unnecessary parenthesis is that it makes yapf slower.

bwbeach:
Replaced with #143.

PR #140, opened 2016-04-28
Fix part size decision for really large files.

bwbeach:
Now that the maximum file size is 10TB, the limit of
10,000 parts per file matters when choosing a part size.


svonohr:
Quick question here: Why does the code try really hard to have all parts larger than the minimum part size? According to the documentation the last part can be smaller.

bwbeach:
That's just me and an excessive preference for symmetry.  You're right that the last part is allowed to be smaller.

PR #146, opened 2016-05-02
Encrypting files for uploads

svonohr:
Replaces #134


bwbeach:
What's the status of this work?  Encryption would be awesome to support.

B2 now supports attaching bucket-info to a bucket, and supports sending the SHA1 of an uploaded file at the end.

svonohr:
This has still some features missing, e.g., storing encryption keys in the bucket info. I don't known when I have time and motivation to work on this.

bwbeach:
I'm still learning how to run an open source project.  What's the right thing to do here?  I'd rather not leave the pull request open.  It looks like we have a branch for this work.  How about closing the pull request and leaving the branch in place?

TJC:
@bwbeach The pull request can be a useful place to record the high level discussion that relates to the branch. Although in that regard, so can an Issue. 

If this pull request is considered a dead end, I would suggest creating a new Issue related to the goal (file encryption support) and then linking from that issue to this pull request, so that people can go back and find this discussion later. Then close the pull request.

bwbeach:
Thanks, @TJC.  I'll do that.

bwbeach:
Closing.  Replaced with issue #387, which has a link to this pull request.

Issue #150, opened 2016-05-07
Provide an "exclude" option for the b2 sync command


segphault:
It'd be really useful to be able to exclude paths that match a certain pattern from synchronization. For example, if I'm synchronizing a bunch of node.js projects, I might want to have b2 sync ignore all the node_modules folders.

bwbeach:
Good idea. 

Do you want the excluded things to be wildcards?  Regexp or glob?  Match the entire path, or just the file name?

I assume that if a folder is excluded, everything under it should be excluded, too.

segphault:
For the case that I care about personally, I want to be able to match just the file name rather than the whole path, but I could see the latter also being useful in some situations. I'd like to be able to do something like `--exclude "node_modules"` and have it exclude directories that match that name everywhere they appear in the hierarchy.

Wildcards would also be great, I think either glob or regex would work well. And yep, it'd also skip the contents of excluded folders. Thanks!

segphault:
I decided to take a crack at implementing this myself. Given the way that the sync feature builds its file list, matching against the full path ended up being the most practical and least intrusive approach. I also decided on using a regex for the match.

sund:
I would also like to see rsync-style exclusion and inclusion, i.e. from the rsync help:

`
--exclude=PATTERN     exclude files matching PATTERN

--exclude-from=FILE     read exclude patterns from FILE

--include=PATTERN     don't exclude files matching PATTERN

--include-from=FILE     read include patterns from FILE
`

Just my thoughts.

ppolewicz:
@sund this seems like a very small change after #153 is merged. Would you like to implement the options you mentioned?

TJC:
Just wanted to say -- I think you should try and implement the same exclude patterns as rsync; they've become reasonably well-understood by many people, and supported by third-party apps.

ppolewicz:
@TJC @sund @segphault do you still need this?

simongreen-net:
I would like to see this option.

ppolewicz:
@simongreen-net okay. Would you like to implement it?

We encourage outside contributors to perform changes on our codebase. Many such changes have been merged already. In order to make it easier to contribute, core developers of this project:

* provide guidance (through the issue reporting system)
* provide tool assisted code review (through the Pull Request system)
* maintain a set of integration tests (run with a production cloud)
* maintain a set of (well over a hundred) unit tests
[...]

PR #153, opened 2016-05-07
Added support for excluding files from sync

segphault:
Implements the feature requested in #150.

### DIFF
--- a/b2/console_tool.py
+++ b/b2/console_tool.py
@@ -546,7 +546,7 @@ class MakeUrl(Command):
 class Sync(Command):
     """
     b2 sync [--delete] [--keepDays N] [--skipNewer] [--replaceNewer] \\
-            [--threads N] [--noProgress] <source> <destination>
+            [--threads N] [--noProgress] [--exclude N] <source> <destination>
 
         Copies multiple files from source to destination.  Optionally
         deletes or hides destination files that the source does not have.
@@ -555,6 +555,11 @@ class Sync(Command):
         number of threads is 10.  Progress is displayed on the
         console unless '--noProgress' is specified.  A list of
         actions taken is always printed.
+
+        You can specify --exclude to selectively ignore files that
+        match the given pattern. Ignored files will not copy during
+        the sync operation. The pattern is a regular expression
+        that is tested against the full path of each file.
 
         Files are considered to be the same if they have the same name
         and modification time.  A future enhancement may add the ability
@@ -592,6 +597,7 @@ class Sync(Command):
     OPTION_FLAGS = ['delete', 'noProgress', 'skipNewer', 'replaceNewer']
     OPTION_ARGS = ['keepDays', 'threads']
     REQUIRED = ['source', 'destination']
+    LIST_ARGS = ['exclude']
     ARG_PARSER = {'keepDays': float, 'threads': int}
 
     def run(self, args):
--- a/b2/console_tool.py
+++ b/b2/console_tool.py
@@ -546,7 +546,7 @@ class MakeUrl(Command):
 class Sync(Command):
     """
     b2 sync [--delete] [--keepDays N] [--skipNewer] [--replaceNewer] \\
-            [--threads N] [--noProgress] [--exclude N] <source> <destination>
+            [--threads N] [--noProgress] [--exclude <regex>] <source> <destination>
 
         Copies multiple files from source to destination.  Optionally
         deletes or hides destination files that the source does not have.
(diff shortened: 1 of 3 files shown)


b2/sync.py
```diff
@@ -775,17 +780,19 @@ def make_folder_sync_actions(source_folder, dest_folder, args, now_millis, repor
...
-            if source_file is not None:
-                reporter.update_compare(1)
-        else:
-            if dest_file is not None:
-                reporter.update_compare(1)
-        for action in make_file_sync_actions(
-            sync_type, source_file, dest_file, source_folder, dest_folder, args, now_millis
-        ):
-            yield action
+        if not any([is_excluded(ex, source_file) or
```
ppolewicz: shouldn't only the source side be excluded? Also, please use a generator expression rather than a list comprehension here - there is no need to allocate a temporary list.

b2/sync.py
```diff
@@ -752,6 +753,8 @@ def make_file_sync_actions(
             if args.delete:
                 yield LocalDeleteAction(dest_file.name, version.id_)
 
+def is_excluded(ex, target_file):
+    return target_file is not None and ex.match(target_file.name)
```
ppolewicz: why would target_file be None?

b2/sync.py
```diff
@@ -775,17 +780,19 @@ def make_folder_sync_actions(source_folder, dest_folder, args, now_millis, repor
...
-                reporter.update_compare(1)
-        for action in make_file_sync_actions(
-            sync_type, source_file, dest_file, source_folder, dest_folder, args, now_millis
-        ):
-            yield action
+        if not any([is_excluded(ex, source_file) or
+                    is_excluded(ex, dest_file) for ex in exclusions]):
 
+            if source_folder.folder_type() == 'local':
+                if source_file is not None:
```
ppolewicz: why protect against `None`?

ppolewicz:
Travis CI tests have failed. Could you please make sure the code you submit passes tests, and that new functionality is tested as well? We rely on those tests quite a bit.

segphault:
When zip_files generates a union of the two trees, it will yield None in cases where the file exists on one side but not the other. Without that check in `is_excluded`, the application would throw an error complaining that it can't access the `name` property on `target_file`.

It's worth noting that in the second case that you pointed out, the code in question predates my patch: https://github.com/Backblaze/B2_Command_Line_Tool/blob/master/b2/sync.py#L779

b2/console_tool.py
```diff
@@ -546,7 +546,7 @@ def run(self, args):
 class Sync(Command):
     """
     b2 sync [--delete] [--keepDays N] [--skipNewer] [--replaceNewer] \\
-            [--threads N] [--noProgress] <source> <destination>
+            [--threads N] [--noProgress] [--exclude N] <source> <destination>
```
ppolewicz: perhaps `[--exclude <regex>]` would be better?

ppolewicz:
You can run `pre-commit.sh` to fix code style automatically.

ppolewicz:
All looks good, except the missing test.

segphault:
Thanks! I added a test case for exclusion. Does that look good?

bwbeach:
Sorry for not commenting earlier.  I was AFK yesterday.

Nice work!

How do you want `--exclude` to interact with `--delete`?  If you exclude a file from the sync, should it get deleted if it exists at the destination?  I think I would say "yes".  You could accomplish this by filtering the files from the source folder before they get zipped with the files from the destination folder.

b2/console_tool.py
bwbeach: How about:  `[--exclude <regexp>]`

bwbeach:
I was debating with myself yesterday whether regexp or glob would be better for pattern matching, and whether excluding a folder name should automatically exclude everything in it.

Regexp is more powerful, but harder to use.  I'm fine with using it.

I might like to automatically exclude files in folders, though.  I think that's a common use case, and making people stick `/.*` on the end of the regexp seems unnecessary.  As it is, saying `--exclude /my/folder` will do nothing, because the regexp won't match `/my/folder/file1.txt`.

ppolewicz:
```python
>>> re.match('foo/', 'foo/bar')
<_sre.SRE_Match object at 0x7fcd4d9c0370>
>>>
```
so sticking `/.*` at the end is not necessary.

As for glob vs. regex, maybe we can use `--exclude-regex` and if someone will want glob later, we can add `--exclude`?

segphault:
Thanks for the feedback! I moved the exclusion check so that it's performed inside of zip_folders on the `all_files` iterator. That appears to make the feature work as expected with the the --delete option. I added an extra test case to specifically check exclusion and deletion together.

I also renamed the option to --excludeRegex as @ppolewicz  suggests so that somebody else can add an --exclude option with glob support at some point later on.

Issue #151, opened 2016-05-07
Add option to sync command to ignore modification time.


kevincox:
I have a use case that sounds pretty common. I am generating a set of files that have a hash in their file name so that any time the files have the same name they will have the same content. Think of rails asset pipeline where a hash is inserted into the file name. I then want to upload all new files to b2 so that they can be used.

The sync command is almost perfect as this except that even if the files have the same name and contents they will be reuploaded because the modification time is newer. It would be nice to have an option that would ignore the modification time and just use the name (and possibly the size).

(Bonus points: add an option to get an error if the size is different rather then reuploading or skipping)

This is similar to rsync's `--ignore-existing` which only copies new files.

ppolewicz:
This doesn't seem like a big change. Will you try to add it yourself?

svonohr:
I might work on that later today, if kevincox doesn't want to.

kevincox:
I have no problem working on it myself but I have a week or two of backlog before I get to this so feel free to take it on @svonohr. If not I'll let you know when I have time to tackle it myself.

svonohr:
What are the exact flags and behavior that we want? Is `--ignore-existing`, which only compares by name, enough? If not, what should the flags be called?

kevincox:
That would definitely be enough to start with. Although I can imagine things like `--assert-size` and `--ignore-size` could also be useful options. I think making "ignore size" the default would make a lot of sense and adding the "assert size" later could be a future improvement.

bwbeach:
Another user has requested matching the sha1 of the file contents, not just the modification time.  We don't need to implement this now, but it might be nice to take it into account when designing the command-line options.

I think that it's safe to assume that sync matches file names, and then uses additional criteria to decide if the file is the "same", or needs to be updated.  Maybe a generic `--compareVersions` option would make sense, with the default value being `modTime`, and other options being `none`, `size`, `sha1`.  I think the option `none` would solve this issue.

Another `--errorIfDifferent` option, with the same set of possible values, would let you verify that the size or sha1 matches for files that should be the same based on `--compareVersions`.

svonohr:
I like these options. It's certainly more structured. Except for sha1, the options should be easy to implement.

PR #154, opened 2016-05-08
Option to ignore modification time

svonohr:
As discussed in #151.
It's still missing tests for the new option.


svonohr:
What needs to be in the doc string?

bwbeach:
The doc string for `sync` should mention the `--compareVersions` option, say what it does, what the three choices are, and what the default is.

svonohr:
I think I did that in the second commit. Or did I miss some doc strings?

bwbeach:
Looks good.  Sorry I missed that.

svonohr:
This should be ready for merging into master.

ppolewicz:
LGTM

PR #158, opened 2016-05-12
Make it so tests can run without touching PYTHONPATH

ppolewicz:
This lets setuptools take care of test dependencies and in the future will allow for supporting Jython and Windows.
As setuptools now handle dependencies, we are less locked in Travis CI (python setup.py nosetests can be triggered by any CI easily).

script is used instead of Makefile for running development tests. Makefile didn't work on Ubuntu environments for some reason.


bwbeach:
Nicely done.

ppolewicz:
https://github.com/nose-devs/nose/issues/974 might have been hit

Issue #152, opened 2016-05-07
b2 upload_file to accept stdin


olcto:
A basic unix way of programs interacting with programs is with stdin and stdout using pipe (|). I have a use case where I want to send a backup of a ZFS snapshot to Backblaze B2 cloud storage with compression and encryption. For example on a unix command line:

`zfs send storage@20160507 | gzip | openssl enc -aes-256-cbc -a -salt | b2 upload_file bucket-name - storage_20160507.gz.ssl`

This has many advantages:
- Local HDD space is saved by not having to save the ZFS snapshot to file
- Compression is leveraged using gzip
- The file is encrypted with openssl before it leaves the computer

I have a fork containing a working example of what I need with the B2 Command Line Tool: https://github.com/olcto/B2_Command_Line_Tool

ppolewicz:
Your code buffers a part in memory. Maybe it is ok, maybe not - I'm just pointing it out for reconsideration.

By the way, there is a branch with encryption enabled natively in b2 CLI. It is being worked on.

olcto:
I agree my implementation is not very memory efficient, especially if `part_size` were to be increased to the maximum allowable 5GB for multipart upload, the 32-bit python interpreter would most certainly crash (untested). Maybe a good solution would be caching the `part_size` to a temporary file? I am open to ideas and ways to implement this.

In my use case I would most likely stay with an external program from b2 CLI for encrypting my files.

olcto:
Just played around with python tempfile library and the data is now chunked into a temporary file. This changes the memory footprint drastically from a couple hundred MBs to about 30 MB and should be invariant to `part_size`.

ppolewicz:
Caching in a temporary file is not resource efficient either.

Properly solving this problem would require b2 API to accept files of unknown size.

@bwbeach, @svonohr - do you have any idea on how we could make this work with keeping resource consumption on a reasonable level?

olcto:
Just to note, the temporary file is only of `part_size` which can range from 100MB to 5GB, not of the entire uploaded file. Seems like a reasonable compromise between HDD space vs RAM usage, especially on memory constrained systems.

ppolewicz:
this is not a good solution for embedded systems. They don't have 5GB of temp space or 100MB of ram to spare.

olcto:
[...]

PR #161, opened 2016-05-16
stdin for b2 upload_file

olcto:
#152

Not ready for merge yet:
- [ ] Needs to be reviewed
- [ ] Create unit tests

### DIFF
--- a/b2/bucket.py
+++ b/b2/bucket.py
@@ -16,7 +16,8 @@ import os
 import io
 
 from .exception import (
-    AlreadyFailed, B2Error, MaxFileSizeExceeded, MaxPartsExceeded, MaxRetriesExceeded, UnrecognizedBucketType
+    AlreadyFailed, B2Error, MaxFileSizeExceeded, MaxPartsExceeded, MaxRetriesExceeded,
+    UnrecognizedBucketType
 )
 from .file_version import FileVersionInfoFactory
 from .progress import DoNothingProgressListener, AbstractProgressListener, RangeOfInputStream, StreamWithProgress
@@ -275,19 +276,13 @@ class Bucket(object):
         upload_source = UploadSourceLocalFile(local_path=local_file, content_sha1=sha1_sum)
         return self.upload(upload_source, file_name, content_type, file_infos, progress_listener)
 
-    def upload_stream(
-        self,
-        file_name,
-        part_size,
-        content_type=None,
-        file_infos=None,
-    ):
+    def upload_stream(self, file_name, part_size, content_type=None, file_infos=None,):
         """
-        Uploads a file from stdin stream to a B2 file.
+        Uploads from stdin stream to a B2 file.
         """
         #determine if it requires large file upload, > part_size >= minimum_part_size
         #TODO make part_size user configurable 100MB to 5GB so max file size of 10TB can be reached with 1 to 10000 parts
-        chunk_size = 16*1024 #16KiB chunk size for memory considerations
+        chunk_size = 16 * 1024  #16KiB chunk size for memory considerations
         validate_b2_file_name(file_name)
         file_info = file_infos or {}
         content_type = content_type or self.DEFAULT_CONTENT_TYPE
@@ -295,25 +290,27 @@ class Bucket(object):
 
         temp_fd, temp_path = tempfile.mkstemp()
         stdin_reader = io.open(sys.stdin.fileno())
-        stdin_buffered = io.BufferedReader(stdin_reader.buffer) #Allows for peeking stdin
-        with os.fdopen(temp_fd, 'w+b') as tmp:
+        stdin_buffered = io.BufferedReader(stdin_reader.buffer)  #Allows for peeking stdin
+        with os.fdopen(temp_fd, 'w+b') as temp_file:
             for i in six.moves.xrange(0, part_size, chunk_size):
                 #checks if end of file chunking to be in line with part_size
                 if (i + chunk_size) > part_size:
                     chunk = stdin_buffered.read(part_size - i)
                 else:
                     chunk = stdin_buffered.read(chunk_size)
-                if not chunk: #EOF
+                if not chunk:  #EOF
                     break
-                tmp.write(chunk)
-            tmp_length = tmp.tell()
-            tmp.seek(0)
+                temp_file.write(chunk)
+            temp_file_length = temp_file.tell()
+            temp_file.seek(0)
             #check if small file upload
             #peek to check if stream ended on first part_size boundary
-            if (tmp_length < part_size) or not stdin_buffered.peek(1):
+            if (temp_file_length < part_size) or not stdin_buffered.peek(1):
                 #print 'Starting small file upload'
                 upload_source = UploadSourceLocalFile(temp_path)
-                return self._upload_small_file(upload_source, file_name, content_type, file_info, progress_listener)
+                return self._upload_small_file(
+                    upload_source, file_name, content_type, file_info, progress_listener
+                )
             else:
                 #print 'Starting large file multipart upload'
                 #TODO Figure out how to incorporate progress listener
@@ -339,30 +336,29 @@ class Bucket(object):
                         raise MaxFileSizeExceeded(large_file_size, self.MAX_LARGE_FILE_SIZE)
 
                     result = self._upload_part(
-                        unfinished_file.file_id,
-                        part_index,
-                        part_range,
-                        upload_source,
+                        unfinished_file.file_id, part_index, part_range, upload_source,
                         large_file_upload_state
                     )
                     part_sha1_array.append(result['contentSha1'])
-                    #clear contents of tmp file
-                    tmp.seek(0)
-                    tmp.truncate()
+                    #clear contents of temp_file file
+                    temp_file.seek(0)
+                    temp_file.truncate()
                     for i in six.moves.xrange(0, part_size, chunk_size):
                         if (i + chunk_size) > part_size:
                             chunk = stdin_buffered.read(part_size - i)
                         else:
                             chunk = stdin_buffered.read(chunk_size)
-                        if not chunk: #EOF
+                        if not chunk:  #EOF
                             break
-                        tmp.write(chunk)
-                    if tmp.tell() == 0:
+                        temp_file.write(chunk)
+                    if temp_file.tell() == 0:
                         #Reached EOF
                         break
-                    tmp.seek(0)
+                    temp_file.seek(0)
                 #Finish the large file
-                response = self.api.session.finish_large_file(unfinished_file.file_id, part_sha1_array)
+                response = self.api.session.finish_large_file(
+                    unfinished_file.file_id, part_sha1_array
+                )
                 #TODO probably check final sha1 of file and sha1 array
                 return FileVersionInfoFactory.from_api_response(response)
 
--- a/b2/bucket.py
+++ b/b2/bucket.py
@@ -10,8 +10,8 @@
 
 import six
 import threading
-from cStringIO import StringIO
 import sys
+import tempfile
 
 from .exception import (
     AlreadyFailed, B2Error, MaxFileSizeExceeded, MaxRetriesExceeded, UnrecognizedBucketType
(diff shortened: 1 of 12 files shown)


ppolewicz:
Please run the `pre-commit.sh` script. It reports some errors and applies some fixes on its own.

b2/bucket.py
```diff
@@ -271,6 +275,97 @@ def upload_local_file(
...
+        chunk_size = 16*1024 #16KiB chunk size for memory considerations
+        validate_b2_file_name(file_name)
+        file_info = file_infos or {}
+        content_type = content_type or self.DEFAULT_CONTENT_TYPE
+        progress_listener = DoNothingProgressListener()
+
+        temp_fd, temp_path = tempfile.mkstemp()
+        stdin_reader = io.open(sys.stdin.fileno())
+        stdin_buffered = io.BufferedReader(stdin_reader.buffer) #Allows for peeking stdin
+        with os.fdopen(temp_fd, 'w+b') as tmp:
```
ppolewicz: `tmp` is not a very descriptive variable name. `temporary_file`

b2/bucket.py
```diff
@@ -271,6 +275,97 @@ def upload_local_file(
...
+                tmp.write(chunk)
+            tmp_length = tmp.tell()
+            tmp.seek(0)
+            #check if small file upload
+            #peek to check if stream ended on first part_size boundary
+            if (tmp_length < part_size) or not stdin_buffered.peek(1):
+                #print 'Starting small file upload'
+                upload_source = UploadSourceLocalFile(temp_path)
+                return self._upload_small_file(upload_source, file_name, content_type, file_info, progress_listener)
+            else:
```
ppolewicz: since the function returns at the end of the `if` block, you could continue the code just after it, which would save the `else:` statement and the indentation that comes after it. That would make the subsequent code easier to read.

b2/bucket.py
```diff
@@ -271,6 +275,97 @@ def upload_local_file(
...
+        """
+        #determine if it requires large file upload, > part_size >= minimum_part_size
+        #TODO make part_size user configurable 100MB to 5GB so max file size of 10TB can be reached with 1 to 10000 parts
+        chunk_size = 16*1024 #16KiB chunk size for memory considerations
+        validate_b2_file_name(file_name)
+        file_info = file_infos or {}
+        content_type = content_type or self.DEFAULT_CONTENT_TYPE
+        progress_listener = DoNothingProgressListener()
+
+        temp_fd, temp_path = tempfile.mkstemp()
```
ppolewicz: if b2 CLI (or a program using the b2 API library) dies for some reason during the upload, the file will leak and consume space in the user system until someone comes and removes the file manually. Using `tempfile.TemporaryFile` would prevent the leak on some systems.

b2/console_tool.py
```diff
@@ -673,10 +673,10 @@ def run(self, args):
 class UploadFile(Command):
     """
     b2 upload_file [--sha1 <sha1sum>] [--contentType <contentType>] [--info <key>=<value>]* \\
-            [--noProgress] [--threads N] <bucketName> <localFilePath> <b2FileName>
+            [--noProgress] [--threads N] [--partSize <partSize>] <bucketName> <localFilePath> <b2FileName>
 
         Uploads one file to the given bucket.  Uploads the contents
-        of the local file, and assigns the given name to the B2 file.
+        of the local file or specify '-' for STDIN, and assigns the given name to the B2 file.
```
ppolewicz: this doesn't read very well. "Uploads the contents of the local file, and assigns the given name to the B2 file. Specify '-' for to use standard input". Man of cat and bash does not use "STDIN" - we shouldn't either.

b2/bucket.py
```diff
@@ -271,6 +273,58 @@ def upload_local_file(
...
+                sha1_array.append(sha1_sum)
+                part_number += 1
+                print 'Part %d of size %d bytes being uploaded ...' %(part_number, len(data))
+                response = self.api.raw_api.upload_part(
+                    upload_url,
+                    upload_auth_token,
+                    part_number,
+                    len(data),
+                    sha1_sum,
+                    StringIO(data)
```
bwbeach: For compatibility between Python 2 and 3, I think you want six.BytesIO here.

b2/bucket.py
```diff
@@ -271,6 +273,58 @@ def upload_local_file(
...
+            unfinished_file = self.start_large_file(file_name, content_type, file_info)
+            part_number = 0
+            sha1_array = []
+            while True:
+                upload_url, upload_auth_token = self._get_upload_part_data(unfinished_file.file_id)
+                sha1_sum = hex_sha1_of_bytes(data)
+                sha1_array.append(sha1_sum)
+                part_number += 1
+                print 'Part %d of size %d bytes being uploaded ...' %(part_number, len(data))
+                response = self.api.raw_api.upload_part(
```
bwbeach: Can you use the existing Bucket._upload_part() method here?  It already knows how to compute the sha1, and has hooks for progress bars.

b2/bucket.py
```diff
@@ -285,41 +285,58 @@ def upload_stream(
...
-                    len(data),
-                    sha1_sum,
-                    StringIO(data)
-                )
-                assert sha1_sum == response['contentSha1']
-                data = sys.stdin.read(part_size)
-                if not data:
-                    print 'Reached end of inputstream'
+        with tempfile.TemporaryFile() as tmp:
+            for i in six.moves.xrange(0, part_size, chunk_size):
```
bwbeach: Does this assume that the part size is a multiple of the chunk size?

b2/bucket.py
```diff
@@ -285,41 +285,58 @@ def upload_stream(
...
+                unfinished_file = self.start_large_file(file_name, content_type, file_info)
+                part_number = 0
+                sha1_array = []
+                while True:
+                    upload_url, upload_auth_token = self._get_upload_part_data(unfinished_file.file_id)
+                    sha1_sum = hex_sha1_of_stream(tmp, tmp_length)
+                    sha1_array.append(sha1_sum)
+                    part_number += 1
+                    print 'Part %d of size %d bytes being uploaded ...' %(part_number, tmp_length)
+                    response = self.api.raw_api.upload_part(
```
bwbeach: Use Bucket._upload_part()?

b2/bucket.py
```diff
@@ -285,41 +285,58 @@ def upload_stream(
...
+                        part_number,
+                        tmp_length,
+                        sha1_sum,
+                        tmp
+                    )
+                    assert sha1_sum == response['contentSha1']
+                    #clear contents of tmp file
+                    tmp.seek(0)
+                    tmp.truncate()
+                    for i in six.moves.xrange(0, part_size, chunk_size):
```
bwbeach: This code is duplicated.  Factor out?

b2/console_tool.py
```diff
@@ -714,13 +719,21 @@ def run(self, args):
         max_workers = args.threads or 10
         self.api.set_thread_pool_size(max_workers)
 
+        #TODO pull out constants for min and max size, 100MB and 5GB
+        part_size_bytes = human2bytes(args.partSize) or human2bytes('100MB') #100MB default
+	if (part_size_bytes < human2bytes('100MB')) or (part_size_bytes > human2bytes('5GB')):
```
bwbeach: This should use the minimum part size returned by the service, not the 100MB constant.

b2/utils.py
```diff
@@ -259,3 +259,29 @@ def format_and_scale_fraction(numerator, denominator, unit):
...
+        'TB': 1e12,
+        'KiB': 2**10,
+        'MiB': 2**20,
+        'GiB': 2**30,
+        'TiB': 2**40
+    }
+
+    init = s
+    num = ""
+    while s and s[0].isdigit() or s[0] == '.':
```
bwbeach: Do you need to allow "." here?  Limiting it to integers might be simpler.

b2/bucket.py
```diff
@@ -283,25 +285,35 @@ def upload_stream(
...
+        temp_fd, temp_path = tempfile.mkstemp()
+        stdin_reader = io.open(sys.stdin.fileno())
+        stdin_buffered = io.BufferedReader(stdin_reader.buffer) #Allows for peeking stdin
+        with os.fdopen(temp_fd, 'w+b') as tmp:
             for i in six.moves.xrange(0, part_size, chunk_size):
-                chunk = sys.stdin.read(chunk_size)
-                if not chunk:
+                #checks if end of file chunking to be in line with part_size
+                if (i + chunk_size) > part_size:
+                    chunk = stdin_buffered.read(part_size - i)
```
bwbeach: Maybe rename "i' to "offset_in_part" for clarity?

b2/bucket.py
olcto: You are looking at an old commit. I did end up refactoring to use _upload_part() in the latest commit.

b2/bucket.py
olcto: It did assume that, I had a fix in the latest commit.

b2/bucket.py
olcto: The latest commit uses _upload_small_file.

ppolewicz:
@olcto: could you please run the pre-commit script again?

jmealo:
@olcto: If you have time to revisit this, I'd love to see this merged!

bwbeach:
Does anybody want to finish up this work?  

I'm cleaning up old pull requests, and will close in a few days if nothing happens.

bb8:
Are there any plans to integrate this? Using a stream to pipe a v.large file (500GB+) is a common use-case in AWS/GCP and its ommission from the B2 CLI is the only thing stopping us from using B2 more widely at the moment.

ppolewicz:
@bb8 that would be an interesting use case for many reasons. Actually I think we've done some work recently that would make it easier to upload large files like that. I'll look into that to see how much work would this need, given that we have more infrastructure for it now.

By the way, could you say how many files you have and what are they? Is it video or raw VM images or something like that?

ppolewicz:
I suspect you might be dealing with sensor data. Is it compressed? How fast is it streaming?

bb8:
@ppolewicz in our particular case, they are compressed VM snapshots, at a daily frequency. The resulting files are often larger than the remaining available space on the host, so it's not feasible to spool the entire file to disk or memory beforehand.

ppolewicz:
@bb8 is there any temporary storage that could be used to cache fragments for uploading? I you have some space (10-20GB would be ideal, but 2GB will be decent and even 200-300MB will be much faster), performance and reliability of the upload operation could be greatly improved. If you have no space whatsoever and want to upload it without any temporary files, a network connection hiccup would break the transfer and since we don't have the data written anywhere, the entire operation would fail. We could also use memory. Also, please tell me, do you upload a snapshot of the exact same machine every day? We may have another improvement in mind.

ppolewicz:
I'm thinking more about it and I have a couple more questions. What are the VMs hosting? Is it a big sql database, or some sort of more static type of content? Also, if there would be a compelling reason to do it, would it be possible to make the stream non-compressed so that b2 cli would take care of the compression?

bb8:
@ppolewicz The VMs are varied, and all backups are encrypted and compressed with GPG before transfer. In our case, the host server always has at least 50GB available space to allow for smooth operation, but that may not be the case for others. We currently use the AWS CLI for streaming backups to S3, and it works great, but having the option to use B2 would be advantageous.

nilayp:
Nilay from Backblaze here. 
While you are waiting for this issue to be implemented, just wanted to make certain you are aware that Backblaze B2 supports the AWS CLI via B2’s S3 compatible API. You should be able to just add a few configuration items to switch a job from AWS S3 to Backblaze B2. Details can be found here: https://help.backblaze.com/hc/en-us/articles/360047779633-Configuring-the-AWS-CLI-for-use-with-B2?mobile_site=true

ppolewicz:
@bb8 I understand. It makes a lot of sense, actually! Thank you for sharing your usage pattern. I'm thinking a few steps ahead and I'd like to understand whether a solution I have in mind would be viable in your case.

If b2cli could see the unencrypted uncompressed data, it could potentially use the information found within the uncompressed stream to avoid re-transfering the data to the B2 server if it is already there (because the operating system part of the VM snapshot is probably the same today as it was yesterday). I understand that you need encryption, compression and gpg signatures, so in the future, b2cli could take care of those for you. This would be the same operation that you do today, except the order would be a bit different. It would reduce network transfer, CPU usage and backup duration for your machines though, so maybe it's worth to consider it.

Does it seem potentially viable or would it not work for some reason (that I am not aware of)?

b2/bucket.py
mlech-reef: The support for Python 2 is dropped already

mjurbanski-reef:
stdin (or from other FIFO streams) upload support has been since implemented in both b2sdk&CLI. Thank you for your support & showing this was needed.

PR #167, opened 2016-05-30
Progress listener refactoring

svonohr:
- Changed `download_file_by_id`/`download_file_by_name` so that they accept a progress listener
- Added unit tests to check progress for downloading files
- Added unit tests to verify that progress listeners are actually closed


svonohr:
Are the upload functions not supposed to close the progess listener when they're done? The sync command relies on not closing them. I changed it so that they do close the progess listener and the little hack in sync is not required anymore.

PR #169, opened 2016-06-01
Fixed raw simulator

svonohr:
The raw simulator was missing the `get_file_info` command. Additionally some of the results returned do not match the B2 API specification. Note that I had to change some unit tests, but I believe their previous expected value was in fact wrong.


bwbeach:
Looks good to me.  Thanks.

PR #170, opened 2016-06-11
Fix max_upload_workers being ignored in B2Api

dpedu:


svonohr:
Shouldn't the default value for `max_upload_workers` in the `__init__` function be set to `1`, to match the previous behavior?

dpedu:
The original behavior was 10 workers, and was changed seemingly inadvertently in bdd810e.

bwbeach:
Yes.  That was my bad.  Thanks for catching it.

PR #179, opened 2016-06-21
Throw exception when some sync actions fail

### LINKED ISSUE #174: Exit status when some files fail syncing
This is a split from #163.
I can confirm that the exit status is 0, even when some of the files couldn't be synced. It can be easily reproduced by trying to sync a file that cannot be read by the current user. If some sync actions failed we could summarize them in a report at the end.

svonohr:
This fixes #174.
Maybe a unit test should be added, but I don't know how to get the global `make_folder_sync_actions` function to return some dummy actions.


b2/sync.py
```diff
@@ -906,15 +907,20 @@ def sync_folders(source_folder, dest_folder, args, now_millis, stdout, no_progre
...
-            sync_executor.submit(action.run, bucket, reporter)
+            future = sync_executor.submit(action.run, bucket, reporter)
+            action_futures.append(future)
             total_files += 1
             total_bytes += action.get_bytes()
         reporter.end_compare(total_files, total_bytes)
 
         # Wait for everything to finish
         sync_executor.shutdown()
+        error_count = len([f for f in action_futures if f.exception() is not None])
```
ppolewicz:
```python
if any(True for f in action_futures if f.exception() is not None):
    raise CommandError('sync is incomplete')
```

ppolewicz:
A unit test is missing for this change

PR #180, opened 2016-07-01
Don't store bucket upload URLs is SQLite.

### LINKED ISSUE #171: Using a lot of threads with sync cause sqlite errors
I'm attempting to use the sync command to upload 1 million files using 200 threads. In the log, I'm seeing database errors that prevents an individual file from being uploaded:

`b2_upload(/Volumes/Library/Data/1m-files/file100352.txt, file100352.txt, 1463692063000): OperationalError('database is locked',) database is locked`

When that thread fails - the file is left out of the sync. It never is re-tried.

In addition, infrequently B2 CLI crashes out completely because auth credentials couldn't be retrieved:

`ERROR: Missing account data: database is locked  Use: b2 authorize_account`

bwbeach:
Fix #171 

This will keep the database from being a bottleneck when
sync-ing lots of files.  And now that b2_get_upload_url
is a free call, it doesn't cost any money to use it when
uploading a single file; it just costs a little time.

### DIFF
--- a/b2/account_info.py
+++ b/b2/account_info.py
@@ -166,10 +166,10 @@ class UploadUrlPool(object):
         """
         with self._lock:
             pair_list = self._pool[key]
-            if len(pair_list) == 0:
-                return (None, None)
-            else:
+            if pair_list:
                 return pair_list.pop()
+            else:
+                return (None, None)
 
     def clear_for_key(self, key):
         with self._lock:
--- a/b2/account_info.py
+++ b/b2/account_info.py
@@ -303,7 +303,7 @@ class SqliteAccountInfo(AbstractAccountInfo):
         """
         )
         # This table is not used any more.  We may use it again
-        # someday if we same upload URLs across invocations of
+        # someday if we save upload URLs across invocations of
         # the command-line tool.
         conn.execute(
             """
--- a/b2/account_info.py
+++ b/b2/account_info.py
@@ -141,6 +141,42 @@ class AbstractAccountInfo(object):
         pass
 
 
+class UploadUrlPool(object):
+    """
+    For each key (either a bucket id or large file id), holds a pool
+    of (url, auth_token) pairs, with thread-safe methods to add and
+    remove them.
+    """
+
+    def __init__(self):
+        self._lock = threading.Lock()
+        self._pool = collections.defaultdict(list)
+
+    def put(self, key, url, auth_token):
+        """
+        Adds the url and auth token to the pool for the given key.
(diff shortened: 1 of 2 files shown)


b2/account_info.py
```diff
@@ -267,6 +302,9 @@ def _create_tables(self, conn):
            );
         """
         )
+        # This table is not used any more.  We may use it again
+        # someday if we same upload URLs across invocations of
```
ppolewicz: `if we same upload URLs` -> `if we reuse same upload URLs`

bwbeach: I meant "save":  "if we save upload URLs across invocations of"

I'll fix it.  Thanks.

b2/account_info.py
```diff
@@ -141,6 +141,42 @@ def clear_large_file_upload_urls(self, file_id):
...
+            pair = (url, auth_token)
+            self._pool[key].append(pair)
+
+    def take(self, key):
+        """
+        Returns (url, auth_token) if one is available, or (None, None) if not.
+        """
+        with self._lock:
+            pair_list = self._pool[key]
+            if len(pair_list) == 0:
```
ppolewicz:
```python
if not pair_list:
```
is [recommended] (http://stackoverflow.com/a/53752/95920)

b2/account_info.py
```diff
@@ -141,6 +141,42 @@ def clear_large_file_upload_urls(self, file_id):
         pass
 
 
+class UploadUrlPool(object):
```
ppolewicz: I noticed that a unit test for this class is missing, but this is a very simple class and I am fine with not having a test for it. It will be integration-tested anyway.

b2/account_info.py
bwbeach: I find this clearer, but will change if you feel strongly.

b2/account_info.py
bwbeach: I debated with myself on that.  The tests of the SqliteAccountInfo exercise this code thoroughly.  On the other hand, the code might change later so that's not the case any more.

b2/account_info.py
ppolewicz: I do. When I read this (python) code, I am wondering why the author is explicitly checking if the length of the pair_list is zero, instead of doing it as it is always done in python.

b2/account_info.py
bwbeach: I'll go ahead and write the test.  It's easier than writing the reason for not writing the test.  :-)

PR #185, opened 2016-07-06
Account info cleanup

ppolewicz:
This is done mostly so that SqliteAccountInfo is separated from StubAccountInfo. This way we can run almost all tests on Jython.

Moving AccountInfo to a separate directory is good anyway - we had a big directory with everything in it and I hope that we'll move other things to subdirectories as well.


bwbeach:
Should `b2/account_info.py` be removed?

The SqliteAccountInfo class is in two places: `b2/account_info.py` and `b2/account_info/sqlite_account_info.py`.

PR #191, opened 2016-07-12
Add --minPartSize option when uploading a single file.

bwbeach:
Elliott at Backblaze is doing some performance testing and wanted this option.  It may be useful to other people, too.

### DIFF
--- a/b2/console_tool.py
+++ b/b2/console_tool.py
@@ -702,7 +702,7 @@ class UploadFile(Command):
 
         By default, the file is broken into as many parts as possible to
         maximize upload parallelism and increase speed.  The minimum that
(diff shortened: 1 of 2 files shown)


b2/console_tool.py
```diff
@@ -736,7 +742,8 @@ def run(self, args):
                 content_type=args.contentType,
                 file_infos=file_infos,
                 sha1_sum=args.sha1,
-                progress_listener=progress_listener,
+                min_part_size=args.minPartSize,
+                progress_listener=progress_listener
```
ppolewicz: missing `,` at the end causes a 3-line diff instead of a 2-line diff (which is cleaner)

Issue #190, opened 2016-07-11
User reports: UnicodeDecodeError: 'ascii' codec can't decode byte


bwbeach:
```
Traceback (most recent call last): 
File "/usr/local/bin/b2", line 9, in <module> 
load_entry_point('b2==0.5.7', 'console_scripts', 'b2')() 
File "build/bdist.freebsd-10.3-STABLE-amd64/egg/b2/console_tool.py", line 870, in main 
File "build/bdist.freebsd-10.3-STABLE-amd64/egg/b2/console_tool.py", line 797, in run_command 
File "build/bdist.freebsd-10.3-STABLE-amd64/egg/b2/console_tool.py", line 621, in run 
File "build/bdist.freebsd-10.3-STABLE-amd64/egg/b2/sync.py", line 912, in sync_folders 
File "build/bdist.freebsd-10.3-STABLE-amd64/egg/b2/sync.py", line 812, in make_folder_sync_actions 
File "build/bdist.freebsd-10.3-STABLE-amd64/egg/b2/sync.py", line 653, in zip_folders 
File "build/bdist.freebsd-10.3-STABLE-amd64/egg/b2/sync.py", line 622, in next_or_none 
File "build/bdist.freebsd-10.3-STABLE-amd64/egg/b2/sync.py", line 639, in <genexpr> 
File "build/bdist.freebsd-10.3-STABLE-amd64/egg/b2/sync.py", line 498, in all_files 
File "build/bdist.freebsd-10.3-STABLE-amd64/egg/b2/sync.py", line 546, in _walk_relative_paths 
File "build/bdist.freebsd-10.3-STABLE-amd64/egg/b2/sync.py", line 546, in _walk_relative_paths 
File "build/bdist.freebsd-10.3-STABLE-amd64/egg/b2/sync.py", line 546, in _walk_relative_paths 
File "build/bdist.freebsd-10.3-STABLE-amd64/egg/b2/sync.py", line 535, in _walk_relative_paths 
File "/usr/local/lib/python2.7/posixpath.py", line 73, in join 
path += '/' + b 
UnicodeDecodeError: 'ascii' codec can't decode byte 0xe2 in position 51: ordinal not in range(128)
```

bwbeach:
The stack trace shows the latest dev version number, and python 2.7.

After experimenting with things, the only way I could get this error from concatenating strings is by having one str and one unicode:

```
>>> '\xff' + u'foo'
Traceback (most recent call last):
[...]

PR #192, opened 2016-07-13
Work on 190

bwbeach:
Add an assertion to help diagnose #190.

### DIFF
--- a/b2/sync.py
+++ b/b2/sync.py
@@ -537,6 +537,12 @@ class LocalFolder(AbstractFolder):
         names = {}  # name to (full_path, relative path)
         dirs = set()  # subset of names that are directories
         for name in os.listdir(dir_path):
+            # We expect listdir() to return unicode if dir_path is unicode.
+            # This assertion is to help track down issue #190.  One theory
+            # is that listdir() is not behaving correctly.  If it were to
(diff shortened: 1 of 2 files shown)


b2/sync.py
```diff
@@ -537,6 +537,12 @@ def _walk_relative_paths(self, prefix_len, dir_path, reporter):
         names = {}  # name to (full_path, relative path)
         dirs = set()  # subset of names that are directories
         for name in os.listdir(dir_path):
+            # We expect listdir() to return unicode if dir_path is unicode.
+            # This assertion is to help track down issue #190.  One theory
+            # is that listdir() is not behaving correctly.  If it were to
+            # return bytes rather than unicode, it's not clear how to decode it.
+            assert isinstance(name, six.text_type)
```
ppolewicz: if `os.listdir` behaves incorrectly, we'll crash the entire program (in the major release). Could we produce a warning instead, to keep the debug help, but not crash?

bwbeach: I guess that for ASCII file names, the code will work even if the assertion fails.  It will decode bytes with the 'ascii' encoding, and come up with the right answer.

Maybe I'll just fail the assertion if there are non-ASCII characters in the file name, and the file name is not unicode.

PR #199, opened 2016-07-26
Hidden marker keeps next oldest version alive

### LINKED ISSUE #197: Flawed sync --keepDays semantics
Currently `b2 sync --keepDays 30` deletes file versions older than 30 days, by looking at the modification time of the file. This is fatally flawed. If I have a file that was last modified a year ago, and I accidentally delete it, and sync, the sync will delete the file (because version.modtime = 1 year, > 30 days), leaving only a "hidden" marker. Even though I asked to keep 30 days of history, I can't undo a mistake I did a minute ago and my data is gone.

I think it should use the upload timestamp instead.

svonohr:
Fix #197.
The code is ugly, but I don't know how to write it more simpler.


b2/sync.py
```diff
@@ -792,15 +792,21 @@ def make_file_sync_actions(
...
         if args.keepDays is not None and sync_type == 'local-to-b2':
             if dest_file.latest_version().action == 'upload':
                 yield B2HideAction(dest_file.name, dest_folder.make_full_path(dest_file.name))
-        # All versions of the destination file are candidates for cleaning
-        dest_versions_to_clean = dest_file.versions
+                dest_versions_to_clean = dest_file.versions[1:]
 
     # Clean up old versions
     if sync_type == 'local-to-b2':
+        skipNextVersion = False
```
ppolewicz: `skip_next_version`

bwbeach:
I have refactored the code.  In simplifying, I have broken @svonohr's change.  Will fix.

bwbeach:
Now that my head's back in the code, I'm going to argue that we need a more general fix.  Not only should hide markers keep the previous version alive.  Uploaded files should, too.

If I have a version uploaded 5 days ago, and one uploaded 40 days ago, and I have keepDays set to 30, I need to keep the file from 40 days ago if I want to restore the state as of 30 days ago.

svonohr:
You are right. It's kind of unintuitive, but this is what should happen.

bwbeach:
I'll make another pass at the code.

test/test_sync.py
```diff
@@ -398,25 +398,61 @@ def test_delete_b2_multiple_versions(self):
...
+        dst_file = b2_file('a.txt', [-(TODAY - 2 * DAY), TODAY - 4 * DAY, TODAY - 6 * DAY])
+        actions = [
+            'b2_delete(folder/a.txt, id_a_8467200000, (hide marker))',
+            'b2_delete(folder/a.txt, id_a_8294400000, (old version))',
+            'b2_delete(folder/a.txt, id_a_8121600000, (old version))'
+        ]
+        self._check_local_to_b2(None, dst_file, FakeArgs(keepDays=1), actions)
+
+    def test_already_hidden_multiple_versions_keep_days_old_delete(self):
+        dst_file = b2_file('a.txt', [-TODAY + 2 * DAY, TODAY - 4 * DAY])
```
svonohr: Stick to `-(TODAY - 2 * DAY)` for consistency.

b2/sync.py
```diff
@@ -761,64 +762,123 @@ def check_file_replacement(source_file, dest_file, args):
...
+        # Both exist.  Transfer only if the two are different.
+        return files_are_different(source_file, dest_file, args)
+
+
+def make_b2_delete_actions(source_file, dest_file, dest_folder, transferred):
+    for (version_index, version) in enumerate(dest_file.versions):
+        keep = (version_index == 0) and (source_file is not None) and not transferred
+        if not keep:
+            note = ''
+            if version.action == 'hide':
```
svonohr: Maybe put this note stuff into separate function?

b2/sync.py
```diff
@@ -761,64 +762,123 @@ def check_file_replacement(source_file, dest_file, args):
...
+                note = '(hide marker)'
+            elif transferred or 0 < version_index:
```
ppolewicz:
```python
elif transferred or version_index != 0:
```
we don't expect version index to be negative and a few lines above, comparison with zero is used already. I think the important fact is whether it is equal to zero or not.

b2/sync.py
```diff
@@ -761,64 +762,123 @@ def check_file_replacement(source_file, dest_file, args):
...
+        return False
+    elif dest_file is None:
+        # Source file exists, but no destination file.  Always transfer.
+        return True
+    else:
+        # Both exist.  Transfer only if the two are different.
+        return files_are_different(source_file, dest_file, args)
+
+
+def make_b2_delete_actions(source_file, dest_file, dest_folder, transferred):
```
ppolewicz: this function is hard to read. What does it do? The `keep = ...` line is very cryptic

b2/sync.py
```diff
@@ -761,64 +762,123 @@ def check_file_replacement(source_file, dest_file, args):
...
+
+        # We assume that the versions are ordered by time, newest first.
+        assert prev_age_days is None or prev_age_days <= age_days
+
+        # Do we need to hide this version?
+        if version_index == 0 and source_file is None and version.action == 'upload':
+            yield B2HideAction(dest_file.name, dest_folder.make_full_path(dest_file.name))
+
+        # Can we start deleting?  Once we start deleting, all older
+        # versions will also be deleted.
```
ppolewicz: This relies on the fact that `dest_file.versions` are sorted by `mod_time` attribute and this is not asserted, nor documented here or in `file_version.py`. We should at least document it thoroughly, because if the ordering changes and this function will not be updated, it can cause data loss. That kind of bug could also be easily missed by a test, which may accidentally keep the ordering after it is changed in the code due to small homogeneous sample test data.

ppolewicz: Ah, it is asserted! My bad.

b2/sync.py
```diff
@@ -761,64 +762,123 @@ def check_file_replacement(source_file, dest_file, args):
...
+        # Source file exists, but no destination file.  Always transfer.
+        return True
+    else:
+        # Both exist.  Transfer only if the two are different.
+        return files_are_different(source_file, dest_file, args)
+
+
+def make_b2_delete_actions(source_file, dest_file, dest_folder, transferred):
+    for (version_index, version) in enumerate(dest_file.versions):
+        keep = (version_index == 0) and (source_file is not None) and not transferred
```
svonohr: I think the inverted logic of this is much more readable: Delete a file it it's not the newest version, there is no source file or a new file is beeing transfered.

bwbeach:
Nice work.  I like it.  Are we ready to merge?

ppolewicz:
Yes

PR #202, opened 2016-07-29
Include more info in the error message when os.listdir() doesn't retu…

bwbeach:
…rn unicode.


ppolewicz:
This exception cannot be caught as others. It should be defined in exception.py

PR #209, opened 2016-08-08
Catch timeouts thrown by requests libary

ppolewicz:


b2/exception.py
```diff
@@ -8,8 +8,26 @@
...
+
+from .utils import camelcase_to_underscore
+
+
+@six.add_metaclass(ABCMeta)
 class B2Error(Exception):
+    @property
+    def prefix(self):
+        """
+        nice auto-generated error message prefix
```
bwbeach: For methods like this, I like to include an example in the doc string.  Then I can see what it does without having to step through the code in my head.

It looks like it will turn "B2MyException" into "Myexception".

PR #210, opened 2016-08-08
Drop support for Python 3.2

bwbeach:


ppolewicz:
https://github.com/Backblaze/B2_Command_Line_Tool/pull/211 is an alternative which lets us keep 3.2 (for now)

bwbeach:
We don't need to drop 3.2 now, @ppolewicz found a way to work with old `requests`

ppolewicz:
Due to pypi not being able to filter `requests` version `2.11.0` from python3.2, we are dropping 3.2 now. A discussion on that subject with `requests` developers: https://github.com/kennethreitz/requests/issues/3507

PR #212, opened 2016-08-10
Change domain name for B2 service from backblaze.com to backblazeb2.com.

bwbeach:
The support for the new domain went live today.  The old domain (with redirects) will continue to work for six months.


ppolewicz:
Why is this change being done?

bwbeach:
Marketing says: "this helps keep b2 related services distinct from our other initiatives".  Amazon does the same thing with amazon.com and amazonaws.com.

PR #213, opened 2016-08-10
Restore original requests requirement: >=2.9.1

bwbeach:
We don't want to specify an exact version, because that increases the probability of incompatibilities with other packages.


ppolewicz:
merge current master please

ppolewicz:
Now github says this PR changes zero LOC

PR #214, opened 2016-08-18
[DO NOT MERGE] test codacy-github integration

ppolewicz:


ppolewicz:
Codacy support explained to me that the tests here passed because of a tradeoff between false positives and accuracy they had to make.  There is no viable workaround for this.

We can use Codacy PR indicator in a limited fashion: if it detects something, then it is a solid issue or something that needs to be suppressed in Codacy UI. If it does not detect an issue though, then we can't be sure that everything is fine - it might have not been reported properly. Still, it is better than nothing (and as far as I can tell, better than other online tools).

PR #219, opened 2016-08-26
sync: handle out-of-order versions

### LINKED ISSUE #216: sync --keepDays cannot handle out-of-order versions
When using `--keepDays`, commit 0cdddf67ef57c19d8e25e59e0970198e55ca87aa iterates over the existing versions with this assertion:

```
        # We assume that the versions are ordered by time, newest first.
        assert prev_age_days is None or prev_age_days <= age_days
```

However, that assertion may not be true if the previous versions were created with `--replaceNewer`; the timestamps may be in any order. You can see this with, for example:

```
# one version in the bucket
echo one >file
b2 sync ./ $bucket

# now the newer version has an older timestamp
echo two >file
touch -d 20160701 file
b2 sync --keepDays 180 --replaceNewer ./ $bucket
```

If you try to sync again, we hit the assertion:

```
$ b2 sync --keepDays 180 --replaceNewer ./ $bucket
  File "/home/peff/local/bin/b2", line 9, in <module>
    load_entry_point('b2==0.6.5', 'console_scripts', 'b2')()
  File "build/bdist.linux-x86_64/egg/b2/console_tool.py", line 879, in main
  File "build/bdist.linux-x86_64/egg/b2/console_tool.py", line 806, in run_command
  File "build/bdist.linux-x86_64/egg/b2/console_tool.py", line 623, in run
  File "build/bdist.linux-x86_64/egg/b2/sync/sync.py", line 179, in sync_folders
  File "build/bdist.linux-x86_64/egg/b2/sync/sync.py", line 120, in make_folder_sync_actions
  File "build/bdist.linux-x86_64/egg/b2/sync/sync.py", line 85, in make_file_sync_actions
  File "build/bdist.linux-x86_64/egg/b2/sync/policy.py", line 99, in get_all_actions
  File "build/bdist.linux-x86_64/egg/b2/sync/policy.py", line 167, in _get_hide_delete_actions
  File "build/bdist.linux-x86_64/egg/b2/sync/policy.py", line 227, in make_b2_keep_days_actions
AssertionError
```

I don't see a quick fix here. The strategy used in `make_b2_keep_days_actions` does depend on traversing in reverse-version order, so sorting the list before the loop is out. I think it actually needs go back to handling the selective deletion of arbitrary versions from the list.

/cc @bwbeach

peff:
The algorithm for deciding which versions to retain with `--keepDays` requires that the versions be handled in reverse-chronological order. However, this may not be the case if a previous version was added with `--replaceNewer`, which would trigger an assertion.

The goal of the original is to walk backwards and flip on a "deleting" flag, but only _after_ we've walked past the most recent version from before `keepDays` (which itself was the state right _at_ `keepDays`, and should be retained). If our times are out of order, though, we can't rely on a simple flag. Instead, we have to record our current notion of this "sentinel" version to keep, and be ready to replace it if we find something newer.

Fixes #216.
/cc @ppolewicz @bwbeach

### DIFF
--- a/b2/sync/policy.py
+++ b/b2/sync/policy.py
@@ -220,36 +220,48 @@ def make_b2_keep_days_actions(
     only the 25-day old version can be deleted.  The 15 day-old version
     was visible 10 days ago.
     """
-    prev_age_days = None
-    deleting = False
+    # most recent version older than keepDays (which we must keep to
+    # have a complete snapshot as of keepDays)
+    sentinel = None
+    sentinel_index = None
     for version_index, version in enumerate(dest_file.versions):
         # How old is this version?
         age_days = (now_millis - version.mod_time) / ONE_DAY_IN_MS
-
-        # We assume that the versions are ordered by time, newest first.
-        assert prev_age_days is None or prev_age_days <= age_days
 
         # Do we need to hide this version?
         if version_index == 0 and source_file is None and version.action == 'upload':
             yield B2HideAction(dest_file.name, dest_folder.make_full_path(dest_file.name))
 
-        # Can we start deleting?  Once we start deleting, all older
-        # versions will also be deleted.
-        if version.action == 'hide':
-            if keep_days < age_days:
-                deleting = True
+        # Is the version too new? If so, we must keep it.
+        if age_days <= keep_days:
+            continue
 
-        # Delete this version
-        if deleting:
+        # Even if this version does become the sentinel, we know that it
+        # will be useless to keep a sentinel that is just a "hide"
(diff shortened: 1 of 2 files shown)


bwbeach:
@peff - 

I'm merging in your first commit here, which refactors the creation of deletion notes.  It's always nice to make the code tidier.

For the rest of the changes, I'm going to close the pull request because we've decided to go with the upload date/time, not the modification date/time, when making the decision about which files to keep.

peff:
@bwbeach Definitely makes sense given the conversation elsewhere. Thanks.

PR #227, opened 2016-09-12
Add a command to delete a file given just the file ID.

bwbeach:
It first gets the file info, to acquire the file name, then
it deletes the version with the given ID.

### DIFF
--- a/b2/console_tool.py
+++ b/b2/console_tool.py
@@ -279,18 +279,31 @@ class DeleteBucket(Command):
 
 class DeleteFileVersion(Command):
     """
-    b2 delete_file_version <fileName> <fileId>
+    b2 delete_file_version [<fileName>] <fileId>
 
         Permanently and irrevocably deletes one version of a file.
+
+        Specifying the fileName is more efficient than leaving it out.
+        If you omit the fileName, it requires an initial query to B2
+        to get the file name, before making the call to delete the
+        file.
     """
 
-    REQUIRED = ['fileName', 'fileId']
+    OPTIONAL_BEFORE = ['fileName']
+    REQUIRED = ['fileId']
 
     def run(self, args):
-        file_info = self.api.delete_file_version(args.fileId, args.fileName)
-        response = file_info.as_dict()
-        self._print(json.dumps(response, indent=2, sort_keys=True))
+        if args.fileName is not None:
+            file_name = args.fileName
+        else:
+            file_name = self._get_file_name_from_file_id(args.fileId)
+        file_info = self.api.delete_file_version(args.fileId, file_name)
+        self._print(json.dumps(file_info.as_dict(), indent=2, sort_keys=True))
         return 0
+
+    def _get_file_name_from_file_id(self, file_id):
+        file_info = self.api.get_file_info(file_id)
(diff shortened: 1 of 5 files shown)


b2/console_tool.py
```diff
@@ -344,6 +344,25 @@ def run(self, args):
         return 0
 
 
+class GetFileInfoAndDeleteFileVersion(Command):
```
ppolewicz: The command name is too verbose. What it really does, is that it deletes stuff, but the name starts with a "get".

Perhaps the `delete_file_version` command should have the `<fileName>` as an optional argument instead of creating another command? It should then be documented that passing the `fileName` argument has strong positive performance impact, so it should be provided whenever possible.

ppolewicz:
If, contrary to the inline comment, the new command is added, then it should also be added to bash completion.

Also, tests are missing for this change.

b2/console_tool.py
bwbeach: The command-line parser we have requires that optional arguments be last.  What we have now is `b2 delete_file_version <fileName> <fileId>`.

I could rename the new command `delete_file_version_by_id`.

ppolewicz: `delete_file_version_by_id` would be better than `get_file_info_and_delete_file_version`, but I feel that extending `delete_file_version` would be better from user point of view (though it makes the implementation slightly less clean)

ppolewicz: I think the custom logic can be contained within `DeleteFileVersion.parse_arg_list` like this:
```python
OPTIONAL = ['fileName', 'fileId']
REQUIRED = []
def parse_arg_list(self, *args, **kwargs):
    args = super(DeleteFileVersion, self).parse_arg_list(*args, **kwargs)
    if args.fileId is None and args.fileName is None:
        return None
    elif args.fileName is None:  # swap the miss-assigned arguments back to their place
        args.fileId = args.fileName
        args.fileName = None
    return args
```

bwbeach: Maybe `delete_file_id` would be more concise, and work better with file completion.

ppolewicz: Good point with the completion. Since we have `download_file_by_id`, so we could have `delete_by_id` / `delete_file_by_id` / `delete_file_id` too.

bwbeach: Another option is to simply make both arguments optional in the arg parsing, and put logic in the command to figure out what's what.

ppolewicz: That is what I was trying to do in [this code snippet](https://github.com/Backblaze/B2_Command_Line_Tool/pull/227#discussion_r78472610)

bwbeach:
In case you're curious, this came up because a customer has a file with `\uffff` in the file name, and is having trouble typing it on the command line.

ppolewicz:
- it is possible to configure the class to provide _optional_before_ arguments with no _required_ arguments
- it is possible to set the same name of an argument to both _required_ and _optional_before_

I suggest that appropriate assertions and tests are added for such cases. We have a custom-built argument parser and we support extensions of both the library layer and CLI through subclassing. I think a fellow developer should be protected from such mistakes by the argument parser implementation.

PR #231, opened 2016-09-15
pre-commit.sh fixes

Tenzer:
I recently started using B2 and thought I would take a crack at helping improving this utility, but I stumbled across a few quirks in the `pre-commit.sh` script, so here's a bunch of fixes that you may or may not like. It's probably easiest to go through the commits one at a time to get the reason for each change.

A thing I have yet to figure out is how to invoke the `quick` option of skipping the integration tests, as it will always collide with how the branch to check against is specified. I was thinking of adding some proper parameter parsing with getopt, but thought it would be a bit much - at least without consulting you guys.

### DIFF
--- a/pre-commit.sh
+++ b/pre-commit.sh
@@ -83,12 +83,7 @@ done
 
 header test_raw_api
 
-if [[ -n "$TEST_ACCOUNT_ID" && -n "$TEST_APPLICATION_KEY" ]]
-then
-    python -m b2.__main__ test_raw_api
-else
-    echo "Skipping because TEST_ACCOUNT_ID and TEST_APPLICATION_KEY are not set"
-fi
+TEST_ACCOUNT_ID="$(head -n 1 ~/.b2_auth)" TEST_APPLICATION_KEY="$(tail -n 1 ~/.b2_auth)" python -m b2.__main__ test_raw_api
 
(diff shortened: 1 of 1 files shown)


pre-commit.sh
```diff
@@ -105,7 +107,7 @@ function run_integration_tests
     fi
 }
 
-if [[ -z "$PYTHON_VIRTUAL_ENVS" ]]
+if [[ -z "${PYTHON_VIRTUAL_ENVS+1}" ]]
```
bwbeach: It would be worth adding a comment here saying what you said in the commit message, including the link to stackexchange.

bwbeach:
These are all good changes.  Thanks!

pre-commit.sh
bwbeach: Looking at StackExchange, I think I like this answer better:

"You can also use `${foobar-}` to get an empty string when foobar is undefined and the value of foobar otherwise (or put any other default value after the -."

According the the bash manual, it should actually be: `${foobar:-}`

ppolewicz: I found [this](http://stackoverflow.com/a/16753536/95920), might be handy

Tenzer: Thanks! I've amended the check and added a comment about it.

Issue #228, opened 2016-09-13
Defaulting to 10 threads bad idea on consumer ADSL


TJC:
In Australia, the vast majority of consumers have ADSL internet connections, with average upstream bandwidth around 0.7 mbit/sec. Yeah, I know, it sucks.

The b2 tool here seems to split big files into 100mbyte chunks, which it then tries to upload in up to ten simultaneous TCP streams.

This is disastrous. For an average Australian internet user, each of those streams will get 0.07 mbit/s each, or 7kbyte/sec. So it will take nearly four hours to complete -- and the b2 servers seem to give up and kick out connections that have lingered for a while.

Perhaps you could set threads to 1 by default, and suggest that users on high-bandwidth connections increase it? Or at least make it much clearer in the README that threads defaults to 10, which may be wildly inappropriate for low-bandwidth users?

Thanks!

svonohr:
Thanks for the report. I also wondered about the parallel uploading strategy a while ago. While it makes sense for lots of small files, where the overhead of connecting to the server is large, uploading large files with multiple threads doesn't increase the upload bandwidth. Instead many upload connections usually kill the internet performance for everyone else using the same internet connection. Maybe the number of parallel uploads should depend on the filesize?

ppolewicz:
I think this is somewhat similar to #200. It is not easy to implement a reliable automated mechanism that adapts to the (potentially dynamic) environment.
In theory, system could auto-diagnose itself and understand whether the current bottleneck is opening connections, the streaming speed, the hdd speed etc. It could also transfer files using a smart strategy, where if the bottleneck is on the network, and the file/part is about to be finished, then another thread is already started, so that the connection could be established in time to take over the network from the previous file (saving a few seconds of network bandwidth). We have not planned on working towards such optimizations (nor even on simple thread count adaptive mechanism) in the nearby future.

If other cloud storage CLIs don't offer such functionality and we could offer it, that could become an easily benchmarkable competitive advantage of B2 as a service.

@TJC: could you suggest an appropriate change to the README so that, at least, the user knows the thread count?

svonohr:
Another idea that came to mind: Do we keep the connection to the server alive for different requests? I believe we don't (or at least I couldn't find any code for it). Since it's just HTTPS it should be possible to keep the connection alive and avoid creating new expensive HTTPS connections. This also should reduce the number of required threads for fast syncing of lots of small files.

ppolewicz:
[...]

PR #233, opened 2016-09-16
Add information about threads and low-bandwidth connections for #228

bwbeach:
Also, add a warning about using --delete.  A customer lost
files by deleting locally, and then syncing to B2 using --delete.


bwbeach:
@TJC's wording was better.  Took that pull request.

PR #244, opened 2016-09-18
More sync options

neube3:
Implemented most of #150 ("excluding files from sync") requests and including-override added. Note that it still lacks rsync-style (include|exclude)-from=FILE

### DIFF
--- a/b2/console_tool.py
+++ b/b2/console_tool.py
@@ -590,7 +590,7 @@ class Sync(Command):
     """
     b2 sync [--delete] [--keepDays N] [--skipNewer] [--replaceNewer] \\
             [--compareVersions <option>] [--threads N] [--noProgress] \\
-            [--excludeRegex <regex>] <source> <destination>
+            [--excludeRegex <regex>] [--includeRegex <regex>] <source> <destination>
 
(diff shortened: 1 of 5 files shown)


bwbeach:
Nice work.

If I'm understanding correctly, `--includeRegex` only makes sense when you have specified `--excludeRegex`.  Would it be a good idea to sanity check the args, and report an error if `--includeRegex` is used alone?

neube3:
Applied the requested changes.

b2/console_tool.py
```diff
@@ -590,7 +590,7 @@ class Sync(Command):
     """
     b2 sync [--delete] [--keepDays N] [--skipNewer] [--replaceNewer] \\
             [--compareVersions <option>] [--threads N] [--noProgress] \\
-            [--excludeRegex <regex>] <source> <destination>
+            [--excludeRegex <regex>] [--includeRegex <regex>] <source> <destination>
```
ppolewicz: `[--excludeRegex <regex> [--includeRegex <regex>]]`

neube3: Applied requested changes; added note to Sync class documentation about --includeRegex requiring --excludeRegex for clarity.

Issue #115, opened 2016-04-10
UnicodeEncodeError: 'ascii' codec can't encode character u'\xb4' 


jcalonso:
I'm having the the following error, on the latest version of the cli tool (0.5.2)

```
jcalonso@my-server:/media/data# b2 sync myData b2:myBucket/myData
Traceback (most recent call last):
  File "/usr/local/bin/b2", line 9, in <module>
    load_entry_point('b2==0.5.2', 'console_scripts', 'b2')()
  File "/usr/local/lib/python2.7/dist-packages/b2/console_tool.py", line 782, in main
    exit_status = ct.run_command(decoded_argv)
  File "/usr/local/lib/python2.7/dist-packages/b2/console_tool.py", line 258, in run_command
    return self.sync(args)
  File "/usr/local/lib/python2.7/dist-packages/b2/console_tool.py", line 736, in sync
    self._print("+ %s" % filename)
  File "/usr/local/lib/python2.7/dist-packages/b2/console_tool.py", line 292, in _print
    print(*args, file=self.stdout)
UnicodeEncodeError: 'ascii' codec can't encode character u'\xb4' in position 44: ordinal not in range(128)

```

bwbeach:
Thanks for the bug report.  I'll be working on sync this week, and will get this fixed.

bwbeach:
Python uses sys.stdout.encoding to decide how to translate unicode to bytes when printing to stdout.  My guess is that your encoding is set to 'ascii', which means that it won't be able to print unicode.  

My environment has 'utf-8' as the default encoding for sys.stdout, and it works for me.

I'm not sure what the right thing to do is.   When the encoding is unicode-capable, things should work.  When it's not, what should the code do?

bwbeach:
@jcalonso - I think that you could get around this by setting LANG to use an encoding that can print the characters in your file names.   Here's a quick experiment I ran:

    $ echo $LANG
    en_US.UTF-8
    $ python
    Python 2.7.11 (default, Mar  7 2016, 13:29:38) 
    >>> import sys ; sys.stdout.encoding
    'UTF-8'
    >>> print u'\u81ea'
    自
    >>> 
    $ export LANG=en_US.ASCII
    $ python
    Python 2.7.11 (default, Mar  7 2016, 13:29:38) 
    >>> import sys ; sys.stdout.encoding
    'US-ASCII'
    >>> print u'\u81ea'
    Traceback (most recent call last):
[...]

PR #246, opened 2016-09-19
Handle unicode in exception messages

bwbeach:
Fix for resurrected #115

### DIFF
--- a/b2/exception.py
+++ b/b2/exception.py
@@ -10,6 +10,7 @@
 
 from abc import ABCMeta
 
+import json
 import six
 
 from .utils import camelcase_to_underscore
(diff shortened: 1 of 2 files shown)


b2/exception.py
```diff
@@ -10,13 +10,28 @@
...
+        """
+        Python 2 does not like it when you pass unicode as the message
+        in an exception.  We like to use file names in exception messages.
+        To avoid problems, if the message has any non-ascii characters in
+        it, they are replaced with backslash-uNNNN
+
+        https://pythonhosted.org/kitchen/unicode-frustrations.html#frustration-5-exceptions
+        """
+        if six.PY2:
+            if 0 < len(args) and isinstance(args[0], six.text_type):
```
ppolewicz: we usually avoid `<`. How about this:
```python
if args and isinstance(args[0], six.text_type):
```

b2/exception.py
```diff
@@ -10,13 +10,28 @@
...
+                args = tuple([json.dumps(args[0])[1:-1]] + list(args[1:]))
```
ppolewicz: there has to be a better way to encode it than using json

b2/exception.py
bwbeach: works for me.

b2/exception.py
bwbeach: This was quick and easy.  

I could simple write the code to replace chars above 127 with a `\uNNNN` code.  Would you like that better?

ppolewicz: I found out how the [json implementation](https://github.com/python/cpython/blob/master/Lib/json/encoder.py) looks like. They even have it optimized in c... So lets keep your current implementation.

PR #253, opened 2016-09-25
Add InMemoryAccountInfo

ppolewicz:


b2/account_info/in_memory.py
```diff
@@ -0,0 +1,91 @@
...
+
+from functools import wraps
+
+
+def _raise_missing_if_result_is_none(function):
+    @wraps(function)
+    def inner(*args, **kwargs):
+        result = function(*args, **kwargs)
+        if result is None:
+            assert function.__name__.startswith('get_')
```
bwbeach: Putting the assertion outside the `if` would catch problems more reliably.

Better yet, what about putting it outside `inner`?

test/test_account_info.py
```diff
@@ -58,85 +60,38 @@ def test_clear(self):
...
-        except OSError:
-            pass
-
-    def test_account_info(self):
-        account_info = self._make_info()
-        account_info.set_auth_data(
-            'account_id', 'account_auth', 'api_url', 'download_url', 100, 'app_key', 'realm'
-        )
+@six.add_metaclass(ABCMeta)
+class AccountInfoBase(object):
```
bwbeach: How about calling it `AccountInfoTestBase` or `AccountInfoTestMixin`?

ppolewicz: If there is a "[tT]est" in the name, *nose* will collect and call it

bwbeach: Too bad.

ppolewicz: Alternative is to set a `__test__` field to `False`, but that gets inherited, so all child classes need to override this field or they will not get their tests called. I feared this will be error prone.

PR #257, opened 2016-10-05
Add delete_file_version to Bucket

ppolewicz:


ppolewicz:
@sondree please confirm this interface is ok

PR #258, opened 2016-10-09
Add logs using external library

ppolewicz:


ppolewicz:
coverage test fails for external reason. I think it is ready for merge. @bwbeach please test it on your end to see how it really works and judge whether adjustment is necessary

bwbeach:
It looks like this change adds a single logging call.  Is that your intention?

ppolewicz:
No. Bad merge. Fixing.

PR #274, opened 2016-10-29
Call for pylint

szepeviktor:
Let's see what pylint has to say.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -223,6 +223,7 @@ script:
             echo skip
         fi
   - if [ "$MODE" == 'SCA' ]; then pyflakes . || exit; else echo skip; fi
+  - if [ "$MODE" == 'SCA' ]; then pip install pylint && pylint b2 || true; else echo skip; fi
 
   # isort doesn't get along with yapf. Hopefully we'll figure it out and re-enable in the future.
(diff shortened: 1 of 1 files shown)


szepeviktor:
In the Travis log: https://travis-ci.org/Backblaze/B2_Command_Line_Tool/jobs/171681078#L408

szepeviktor:
We have a permissive `|| true`.

ppolewicz:
We run pylint via Codacy, which aggregates various checkers from various pylint-like tools. It also allows us to store suppresions on their server, so our code is not littered with it.

Default pylint configuration is horrible. It spews out garbage like "not enough methods in a class" and "missing docstring". We have adjusted the configuration in Codacy accordingly and we use it on regular basis. For code formatting we use yapf.

I am against merging this and running `pylint` twice, so that one result can be ignored. Shall we close it?

szepeviktor:
OK, I see.
https://www.codacy.com/app/b2_command_line_tool/B2_Command_Line_Tool/issues/ignored

szepeviktor:
Not even this *error*?
`console_tool.py:862,70: Class 'Command' has no '__subclasses__' member (no-member)`

ppolewicz:
No.

PR #278, opened 2016-10-30
Included sync command --dryRun switch in README.md

magnetikonline:
As above.

And just to add - this new `--dryRun` switch works perfectly!

  Pete


ppolewicz:
@bwbeach ready for merge

PR #286, opened 2016-11-10
Avoid UnicodeEncodeError on improperly configured terminals

ppolewicz:
I had some trouble in configuring my terminal improperly, so I was not able to test it. Is there a reliable way someone could recommend?

### DIFF
--- a/b2/console_tool.py
+++ b/b2/console_tool.py
@@ -151,28 +151,29 @@ class Command(object):
         )
 
     def _print(self, *args, **kwargs):
-        try:
-            print(*args, file=self.stdout, **kwargs)
-        except UnicodeEncodeError:
-            print(
-                "\nERROR: Unable to print unicode.  Encoding for stdout is: '%s'" %
-                (sys.stdout.encoding,),
-                file=sys.stderr
-            )
-            print("Trying to print: %s" % (repr(args),), file=sys.stderr)
-            sys.exit(1)
+        self._print_helper(self.stdout, self.stdout.encoding, 'stdout', *args, **kwargs)
 
     def _print_stderr(self, *args, **kwargs):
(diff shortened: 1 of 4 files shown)


ppolewicz:
not yet ready for merge - no test coverage

ppolewicz:
@svonohr @bwbeach do you have a good idea for a test for this? I tried to mis-configure my environment, but for some reason it still worked. My terminal may be too flexible.

bwbeach:
What are you wanting to test?

You could unit test the code by replacing sys.stdout with something that raises a UnicodeEncodeError.  But I'm not worried about the code at that level.

Printing non-ascii raises the exception on my terminal:

```$ LANG=en_US.ASCII python
Python 3.5.2 (default, Aug  8 2016, 11:17:15) 
[GCC 4.2.1 Compatible Apple LLVM 7.3.0 (clang-703.0.31)] on darwin
Type "help", "copyright", "credits" or "license" for more information.
>>> import sys
>>> sys.stdout.encoding
'US-ASCII'
>>> print('\u1234')
  File "<stdin>", line 1, in <module>
UnicodeEncodeError: 'ascii' codec can't encode character '\u1234' in position 0: ordinal not in range(128)
>>> 
```

I'm in favor of just merging this change without a test.

bwbeach:
It works fine when I run it by hand:

```$ LANG=en_US.ASCII python -m b2 ls --versions brianb_20151009 backup

WARNING: Unable to print unicode.  Encoding for stdout is: 'US-ASCII'
Trying to print: ('backup/\u81ea\u7531.txt',)
b'backup/\\u81ea\\u7531.txt'
```

ppolewicz:
ready for merge

Issue #283, opened 2016-11-03
Broken pipe


brianlukoff:
I'm getting this error frequently when using `b2 sync` on a large number of files:

MaxRetriesExceeded() FAILED to upload after 5 tries. Encountered exceptions: Broken pipe: unable to send entire request

ppolewicz:
What is your upload speed? What is the file size? How many treads are you using?

brianlukoff:
They are large files, but I'm on a 1Gbps connection so that shouldn't be an issue - but I'm only getting about 70KB/s (most files are skipped, though, since they are already uploaded). I'm using the default number of threads (10?).

bwbeach:
It's possible that these are because of some operational issues we've had this week with the B2 service.  We'll be deploying some bug fixes in a few hours (2pm PDT) that may help.

If you can run with the `--debugLogs` option, that might give us some useful information for finding the problem.

brianlukoff:
`--debugLogs` is not recognized as a valid option in my version of `b2`.

I am still receiving the same errors when running `b2 sync`.

ppolewicz:
Could you please upgrade to a newer version of b2 and then run again with `--debugLogs`?

brianlukoff:
OK -- didn't realize a newer version was available:

```
$ time b2 sync --keepDays 30 --debugLogs --replaceNewer /Volumes/... b2://.../
b2_upload(..., 1309818392000): MaxRetriesExceeded() FAILED to upload after 5 tries. Encountered exceptions: Broken pipe: unable to send entire request
Broken pipe: unable to send entire request
Broken pipe: unable to send entire request
Broken pipe: unable to send entire request
Broken pipe: unable to send entire request
b2_upload(... 1329019804000): MaxRetriesExceeded() FAILED to upload after 5 tries. Encountered exceptions: Broken pipe: unable to send entire request
Broken pipe: unable to send entire request
Broken pipe: unable to send entire request
Broken pipe: unable to send entire request
Broken pipe: unable to send entire request
b2_upload(..., 1300943123000): MaxRetriesExceeded() FAILED to upload after 5 tries. Encountered exceptions: Broken pipe: unable to send entire request
Broken pipe: unable to send entire request
Connection error: ('Connection aborted.', error(41, 'Protocol wrong type for socket'))
Broken pipe: unable to send entire request
Broken pipe: unable to send entire request
```

ppolewicz:
[...]

PR #287, opened 2016-11-10
Various logging improvements for #283

ppolewicz:


b2/bucket.py
```diff
@@ -370,6 +373,7 @@ def _upload_small_file(
                     return FileVersionInfoFactory.from_api_response(upload_response)
 
             except B2Error as e:
+                logger.exception('error when uploading, token was %s', upload_auth_token)
```
bwbeach: Don't log the auth token.  That *is* sensitive.  It's the URL that's interesting for finding the logs on the server side.

ppolewicz: Nice catch

PR #295, opened 2016-12-09
New bucket fields

### LINKED ISSUE #294: support for bucket info and lifecycle rules
The B2 service now supports storing bucket info with a bucket, and will soon support setting lifecycle rules to automatically delete old versions of files.

Any thoughts on what the interface to the command-line tool should be?

The obvious thing to do is to add `--bucket-info` and `--lifecycle-rules` options to the `create_bucket` and `update_bucket` operations.  Each would take JSON as an argument, like this:

```
b2 update-bucket --bucket-info '{"color": "blue"}' my-bucket
```

That seems reasonable, but may be inconvenient for storing large bucket info or a long list of lifecycle rules.  We could support pulling from a file, perhaps by using the curl convention of "@" to indicate a file.  (That will be unambiguous because JSON can't start with "@".)

```
b2 update-bucket --lifecycle-rules @rules.json my-bucket
```

Does this sound reasonable?

bwbeach:
Fix #294


b2/bucket.py
```diff
@@ -99,19 +99,38 @@ class Bucket(object):
...
         self.id_ = id_
         self.name = name
         self.type_ = type_
+        self.bucket_info = bucket_info or {}
+        self.revision = revision
 
     def get_id(self):
         return self.id_
 
+    def set_info(self, new_bucket_info, ifRevisionIs=None):
```
ppolewicz: `ifRevisionIs` is inconsistent with `if_revision_is` used below.

b2/console_tool.py
```diff
@@ -273,15 +281,29 @@ def run(self, args):
 
 class CreateBucket(Command):
     """
-    b2 create_bucket <bucketName> [allPublic | allPrivate]
+    b2 create_bucket [--bucketInfo <jsonOrFile>] [--lifecycleRules <jsonOrFile] <bucketName> [allPublic | allPrivate]
```
ppolewicz: missing `>`

b2/console_tool.py
```diff
@@ -746,17 +766,29 @@ def run(self, args):
 
 class UpdateBucket(Command):
     """
-    b2 update_bucket <bucketName> [allPublic | allPrivate]
+    b2 update_bucket [--bucketInfo <jsonOrFile>] [--lifecycleRules <jsonOrFile] <bucketName> [allPublic | allPrivate]
```
ppolewicz: missing `>`

b2/bucket.py
```diff
@@ -99,19 +99,38 @@ class Bucket(object):
...
+        account_id = self.api.account_info.get_account_id()
+        return self.api.session.update_bucket(
+            account_id, self.id_, bucket_info=new_bucket_info, ifRevisionIs=ifRevisionIs
+        )
+
     def set_type(self, type_):
         account_id = self.api.account_info.get_account_id()
         return self.api.session.update_bucket(account_id, self.id_, type_)
 
+    def update(self, bucket_type=None, bucket_info=None, lifecycle_rules=None, if_revision_is=None):
```
ppolewicz: I suggest that `set_type` and `set_info` use `update` internally, rather than directly interacting with `self.api.session.update_bucket`. It prevents `account_id = self.api.account_info.get_account_id()` from being copied all over the place, which may not seem like much now, but later when any changes are done to `update_bucket()`, it'd be nice to only need to change one area in `Bucket`.

test/test_console_tool.py
```diff
@@ -121,13 +127,72 @@ def test_buckets(self):
...
+    def test_bucket_info_from_file(self):
+
+        self._authorize_account()
+        self._run_command(['create_bucket', 'my-bucket', 'allPublic'], 'bucket_0\n', '', 0)
+
+        with TempDir() as temp_dir:
+            bucket_info = {'color': 'blue'}
+            bucket_info_file = self._make_local_file(temp_dir, 'bucket_info.json')
+            with open(bucket_info_file, 'wb') as f:
+                f.write(six.b(json.dumps(bucket_info)))
```
ppolewicz:
```python
json.dump(bucket_info, f)
```

test/test_console_tool.py
```diff
@@ -121,13 +127,72 @@ def test_buckets(self):
...
+                }
+                '''
+            self._run_command(
+                [
+                    'update_bucket', '--bucketInfo', '@' + bucket_info_file, 'my-bucket',
+                    'allPrivate'
+                ], expected_stdout, '', 0
+            )
+
+    def test_bucket_info_from_json(self):
```
ppolewicz: this test is almost an exact copy of the one above it. Most of those two should be extracted into a private helper function and then concrete tests should be run by parameters. Copied tests increase maintenance cost and should be avoided.

test/test_console_tool.py
```diff
@@ -121,13 +127,72 @@ def test_buckets(self):
...
+                "bucketType": "allPrivate",
+                "lifecycleRules": [],
+                "revision": 2
+            }
+            '''
+        self._run_command(
+            ['update_bucket', '--bucketInfo', json.dumps(bucket_info), 'my-bucket', 'allPrivate'],
+            expected_stdout, '', 0
+        )
+
```
ppolewicz: There are no tests for failures. The test that checks whether the Conflict exception works correctly is especially important as we do know `b2_fuse` project will build a hard consistency model over it.

test/test_console_tool.py
bwbeach: I haven't exposed that functionality on the command line yet.  Whet it's exposed, there should be a test.

test/test_console_tool.py
bwbeach: I haven't exposed that functionality on the command line yet. Whet it's exposed, there should be a test.

ppolewicz: The `b2_fuse` project will use it over the (so far supported) API level as soon as it is released, which may be before command line support is added. Therefore I think this should be tested with corner-case scenarios on the API level before it is merged.

We know about `b2_fuse` and we can synchronize with their developers, but there may be other projects that can start using the API layer at any point without discussing it with us, so we must be careful for that, or there is a risk... Which in this particular case can quite easily lead to data loss.

svonohr:
Maybe we should show a warning when the user tries to change the bucket info using the CLI? Without it they may overwrite their master key for encryption or other important things.

ppolewicz:
@svonohr when the encryption feature will be continued, additional measures for protection of `update_bucket` should be provided: if the user wants to change the content of `ENCRYPTION_KEY` or delete the field entirely, he will have to pass another flag (`--disable-master-key-protection` in ConsoleTool and `disable_master_key_protection=True` in API layer). For now, those don't make any sense, so I'd delay implementing that to encryption branch.

Would that be a suitable solution?

svonohr:
I don't know if we have to make this special for encryption. Maybe a general message informing the user about the importance of the bucket info and the old value he's about to overwrite would be enough.

ppolewicz:
In my opinion it should be special for encryption.

User should be responsible to some extent for using `update_bucket` and the console interface cannot really accommodate warnings.

bwbeach:
I'm going to go ahead and merge this pull request.  Let's continue the discussion about protecting the keys as part of the encryption project.

PR #303, opened 2016-12-16
Check for clock skew between local host and server

### LINKED ISSUE #301: Detect when local clock differs from B2 clock
It would be nice if the command-line tool detected when the clock in the B2 service is too far off from the local clock.  This matters when setting `src_last_modified_millis` in a file upload.  

What should happen when the clocks don't match?  Is printing a warning enough?  Should sync stop?

bwbeach:
Fix #301

### DIFF
--- a/b2/b2http.py
+++ b/b2/b2http.py
@@ -117,6 +117,43 @@ class ResponseContextManager(object):
         self.response.close()
 
 
+class HttpCallback(object):
+    """
+    A callback object that does nothing.  Override pre_request
+    and/or post_request as desired.
+    """
+
+    def pre_request(self, method, url, headers):
+        """
+        Called before processing an HTTP request.
+
+        Returns True if the request should be processed, and False
+        if the request should be aborted.
+
+        Raises an exception if this request should not be processed.
+        The exception raised must inherit from B2Error.
+
+        :param method: One of: 'POST', 'GET', etc.
+        :param url: The URL that will be used.
+        :param headers: The header sent with the request.
+
+        """
+
+    def post_request(self, method, url, headers, response):
+        """
+        Called after processing an HTTP request.
+        Should not raise an exception.
+
+        Raises an exception if this request should be treated as failing.
+        The exception raised must inherit from B2Error.
+
+        :param method: One of: 'POST', 'GET', etc.
+        :param url: The URL that will be used.
+        :param headers: The header sent with the request.
+        :param response: A response object from the requests library.
+        """
+
+
 class B2Http(object):
     """
     A wrapper for the requests module.  Provides the operations
@@ -138,7 +175,7 @@ class B2Http(object):
             ...
     """
 
-    def __init__(self, requests_module=None, after_request_hook=None):
+    def __init__(self, requests_module=None):
         """
         Initialize with a reference to the requests module, which makes
         it easy to mock for testing.
@@ -148,7 +185,13 @@ class B2Http(object):
         """
         requests_to_use = requests_module or requests
         self.session = requests_to_use.Session()
-        self.after_request_hook = after_request_hook
+        self.callbacks = []
+
+    def add_callback(self, callback):
+        """
+        Adds a callback that inherits from HttpCallback.
+        """
+        self.callbacks.append(callback)
 
     def post_content_return_json(self, url, headers, data, try_count=1, post_params=None):
         """
@@ -174,8 +217,10 @@ class B2Http(object):
         # rewind the data back to the beginning.
         def do_post():
             data.seek(0)
+            self._run_pre_request_hooks('POST', url, headers)
             response = self.session.post(url, headers=headers, data=data)
-            return self._run_after_request_hook(response)
+            self._run_post_request_hooks('POST', url, headers, response)
+            return response
 
         response = _translate_and_retry(do_post, try_count, post_params)
 
@@ -234,16 +279,21 @@ class B2Http(object):
 
         # Do the HTTP GET.
         def do_get():
+            self._run_pre_request_hooks('GET', url, headers)
             response = self.session.get(url, headers=headers, stream=True)
-            return self._run_after_request_hook(response)
+            self._run_post_request_hooks('GET', url, headers, response)
+            return response
 
         response = _translate_and_retry(do_get, try_count, None)
         return ResponseContextManager(response)
 
-    def _run_after_request_hook(self, response):
-        if self.after_request_hook is not None:
-            self.after_request_hook(response)
-        return response
+    def _run_pre_request_hooks(self, method, url, headers):
+        for callback in self.callbacks:
+            callback.pre_request(method, url, headers)
+
+    def _run_post_request_hooks(self, method, url, headers, response):
+        for callback in self.callbacks:
+            callback.post_request(method, url, headers, response)
 
 
 def test_http():
--- a/b2/b2http.py
+++ b/b2/b2http.py
@@ -10,6 +10,7 @@
 
 from __future__ import print_function
 
+import arrow
 import logging
 import json
 import socket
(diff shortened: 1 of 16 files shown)


b2/console_tool.py
```diff
@@ -1036,6 +1038,53 @@ def _setup_logging(self, args, command, argv):
...
+
+MONTHS = ['Jan', 'Feb', 'Mar', 'Apr', 'May', 'Jun', 'Jul', 'Aug', 'Sep', 'Oct', 'Nov', 'Dec']
+
+
+def clock_skew_hook(http_response):
+    """
+    Raises an exception if the clock in the server is too different from the
+    clock on the local host.
+
+    The Date header contains a string that looks like: "Fri, 16 Dec 2016 20:52:30 GMT".
```
bwbeach: Is there a better way to parse dates from HTTP headers?

b2/console_tool.py
```diff
@@ -1040,6 +1042,53 @@ def _setup_logging(self, args, command, argv):
...
+    # Check the difference.  The timedelta.total_seconds() method is not available
+    # in Python 2.6, so we'll compute it using the formula from the Python docs.
+    max_allowed = 10 * 60  # ten minutes, in seconds
+    skew = local_time - server_time
+    skew_seconds = int(
+        (skew.microseconds + (skew.seconds + skew.days * 24 * 3600) * 1000000) / 1000000
+    )
+    if max_allowed < abs(skew_seconds):
+        if skew_seconds < 0:
+            raise ClockSkew('local clock is %d seconds behind server' % (-skew_seconds,))
```
ppolewicz: the formatting of this message should not be here, but in `ClockSkew` class, the same as with other exceptions we have.

b2/console_tool.py
```diff
@@ -1040,6 +1042,53 @@ def _setup_logging(self, args, command, argv):
...
+
+    The Date header contains a string that looks like: "Fri, 16 Dec 2016 20:52:30 GMT".
+    The strptime function uses the current locale for month and day names like
+    "Dec" and "Fri", so we convert them to numbers ourselves.
+    """
+    # Make a string that uses month numbers instead of month names
+    server_date_str = http_response.headers['Date']
+    match = DATE_PATTERN.match(server_date_str)
+    if match is None:
+        raise DateFormatBad('date from server is: ' + server_date_str)
```
ppolewicz: the formatting of this message should not be here, but in `ClockSkew` class, the same as with other exceptions we have.

b2/console_tool.py
```diff
@@ -1055,7 +1099,10 @@ def decode_sys_argv():
 
 def main():
     info = SqliteAccountInfo()
-    b2_api = B2Api(info, AuthInfoCache(info))
+    b2_http = B2Http()
```
ppolewicz: I think the API user should have `B2Http` constructor attach `ClockSkewHook()` by default. Then it could be disabled explicitly. As it is now, other API users will not get protected unless they look here and decide to attach the hook explicitly.

b2/b2http.py
```diff
@@ -117,6 +117,43 @@ def __exit__(self, exc_type, exc_val, exc_tb):
...
+    """
+    A callback object that does nothing.  Override pre_request
+    and/or post_request as desired.
+    """
+
+    def pre_request(self, method, url, headers):
+        """
+        Called before processing an HTTP request.
+
+        Returns True if the request should be processed, and False
```
ppolewicz: this paragraph is no longer needed

b2/b2http.py
```diff
@@ -117,6 +117,43 @@ def __exit__(self, exc_type, exc_val, exc_tb):
...
+        if the request should be aborted.
+
+        Raises an exception if this request should not be processed.
+        The exception raised must inherit from B2Error.
```
ppolewicz: It should inherit from `B2HttpCallbackPreRequestException`, which should inherit from `B2HttpCallbackException`, which should inherit from `B2Error` or `B2SimpleError`

b2/b2http.py
```diff
@@ -117,6 +117,43 @@ def __exit__(self, exc_type, exc_val, exc_tb):
...
+
+        """
+
+    def post_request(self, method, url, headers, response):
+        """
+        Called after processing an HTTP request.
+        Should not raise an exception.
+
+        Raises an exception if this request should be treated as failing.
+        The exception raised must inherit from B2Error.
```
ppolewicz: It should inherit from `B2HttpCallbackPostRequestException`, which should inherit from `B2HttpCallbackException`, which should inherit from `B2Error` or `B2SimpleError`

b2/b2http.py
```diff
@@ -117,6 +117,43 @@ def __exit__(self, exc_type, exc_val, exc_tb):
...
+        :param method: One of: 'POST', 'GET', etc.
+        :param url: The URL that will be used.
+        :param headers: The header sent with the request.
+
+        """
+
+    def post_request(self, method, url, headers, response):
+        """
+        Called after processing an HTTP request.
+        Should not raise an exception.
```
ppolewicz: this is contradicted in the next paragraph

b2/console_tool.py
```diff
@@ -1040,6 +1042,48 @@ def _setup_logging(self, args, command, argv):
...
+        # Check the difference.  The timedelta.total_seconds() method is not available
+        # in Python 2.6, so we'll compute it using the formula from the Python docs.
+        max_allowed = 10 * 60  # ten minutes, in seconds
+        skew = local_time - server_time
+        skew_seconds = int(
+            (skew.microseconds + (skew.seconds + skew.days * 24 * 3600) * 1000000) / 1000000
+        )
+        if max_allowed < abs(skew_seconds):
+            if skew_seconds < 0:
+                raise ClockSkew('local clock is %d seconds behind server' % (-skew_seconds,))
```
ppolewicz: Formatting of this message should be in `b2.exceptions` as it is done for all other exceptions.

test/test_console_tool.py
```diff
@@ -580,6 +581,41 @@ def test_bad_terminal(self):
         console_tool = ConsoleTool(self.b2_api, stdout, stderr)
         console_tool.run_command(['b2', 'authorize_account', 'my-account', 'good-app-key'])
 
+    def test_clock_skew_hook_bad_format(self):
+        response = mock.MagicMock()
+        response.headers = {'Date': 'bad format'}
+        with self.assertRaises(BadDateFormat):
+            ClockSkewHook().post_request('POST', 'http://foo.com', {}, response)
```
ppolewicz: Use `example.com` or `example.org`, which are designated for this. `foo.com` is an actual domain that someone has, not sure if we want to associate with them.

b2/console_tool.py
ppolewicz: This is a great idea. Maybe lets call it `clock_antiskew_hook` though? `clock_skew_hook` sounds like something that has a purpose of making sure the clock is properly skewed...

b2/b2http.py
bwbeach: Good idea.

test/test_b2http.py
```diff
@@ -195,3 +196,40 @@ def test_get_content(self):
...
         self.session.get.assert_called_with(self.URL, headers=self.EXPECTED_HEADERS, stream=True)
         self.response.close.assert_called_with()
+
+
+class TestClockSkewHook(TestBase):
+    def test_bad_format(self):
+        response = MagicMock()
+        response.headers = {'Date': 'bad format'}
+        with self.assertRaises(BadDateFormat):
+            ClockSkewHook().post_request('POST', 'http://foo.com', {}, response)
```
ppolewicz: there are several `foo.com` instances left in this class, please change it to `example.com`

b2/b2http.py
```diff
@@ -138,13 +207,25 @@ class B2Http(object):
...
         it easy to mock for testing.
+
+        The optional after_request_hook is called on the Response
+        object after every request that doesn't throw an exception.
         """
         requests_to_use = requests_module or requests
         self.session = requests_to_use.Session()
+        self.callbacks = []
+        if install_clock_skew_hook:
+            self.add_callback(ClockSkewHook())
```
ppolewicz: this line is not covered by unit tests

b2/exception.py
```diff
@@ -116,6 +120,41 @@ def __str__(self):
...
+    """
+
+    def __init__(self, clock_skew_seconds):
+        """
+        :param clock_skew_seconds: The different: local_clock - server_clock
+        """
+        super(ClockSkew, self).__init__()
+        self.clock_skew_seconds = clock_skew_seconds
+
+    def __str__(self):
```
ppolewicz: no UT for that method

PR #307, opened 2016-12-20
Limit the number of sync actions queued in memory.

### LINKED ISSUE #281: Limit sync action queue length to avoid memory explosion
found in #279

bwbeach:
Fix #281

### DIFF
--- a/b2/sync/sync.py
+++ b/b2/sync/sync.py
@@ -174,7 +174,7 @@ class BoundedQueueExecutor(object):
     released when an action finishes.
     """
 
-    def __init__(self, executor, queue_limit=100):
+    def __init__(self, executor, queue_limit):
         self.executor = executor
         self.semaphore = threading.Semaphore(queue_limit)
 
--- a/b2/sync/sync.py
+++ b/b2/sync/sync.py
@@ -163,7 +163,7 @@ def count_files(local_folder, reporter):
     reporter.end_local()
 
 
-class BoundedQueueThreadPoolExecutor(object):
+class BoundedQueueExecutor(object):
     """
     Wraps a futures.Executor and limits the number of requests that
     can be queued at once.  Requests to submit() tasks block until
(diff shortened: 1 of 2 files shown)


b2/sync/sync.py
```diff
@@ -183,7 +218,12 @@ def sync_folders(
         # not the same as the executor in the API object, which is used for
         # uploads.  The tasks in this executor wait for uploads.  Putting them
         # in the same thread pool could lead to deadlock.
-        sync_executor = futures.ThreadPoolExecutor(max_workers=max_workers)
+        #
+        # We use an executor with a bounded queue to avoid using up lots of memory
+        # when syncing lots of files.
+        unbounded_executor = futures.ThreadPoolExecutor(max_workers=max_workers)
+        queue_limit = max_workers + 1000
```
ppolewicz: here the default `queue_limit` is >1000 and in the constructor default is more like 100. Perhaps we could remove the default from the constructor and force the user of this class to select the limit explicitly?

test/test_sync.py
```diff
@@ -689,5 +696,78 @@ def _check_one_file(self, src_type, src_file, dst_type, dst_file, args, expected
...
+        """
+        Makes sure that no more than queue_limit workers are
+        running at once, which checks that the semaphore is
+        acquired before submitting an action.
+        """
+        raw_executor = futures.ThreadPoolExecutor(2)
+        bounded_executor = BoundedQueueExecutor(raw_executor, 1)
+        assert_equal = self.assertEqual
+
+        class CountAtOnce():
```
ppolewicz: this is an old-style class. Should always inherit from `object`.

test/test_sync.py
```diff
@@ -689,5 +696,78 @@ def _check_one_file(self, src_type, src_file, dst_type, dst_file, args, expected
...
+    def test_run_more_than_queue_size(self):
+        """
+        Makes sure that the executor will run more jobs that the
+        queue size, which ensures that the semaphore gets released,
+        even if an exception is thrown.
+        """
+        raw_executor = futures.ThreadPoolExecutor(1)
+        bounded_executor = BoundedQueueExecutor(raw_executor, 5)
+
+        class Counter():
```
ppolewicz: this is an old-style class. Should always inherit from `object`.

PR #325, opened 2017-02-15
Minor improvement for retrying uploads.

bwbeach:
Instead of sleeping and trying again with the same upload URL,
now it immediately gets a new upload URL and tries again.

### DIFF
--- a/b2/b2http.py
+++ b/b2/b2http.py
@@ -83,7 +83,7 @@ def _translate_errors(fcn, post_params=None):
         raise UnknownError(repr(e))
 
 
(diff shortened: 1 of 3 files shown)


bwbeach:
I'm not sure this is worthwhile.  Now that I look at the sleep times, they're not very long.

bwbeach:
Upon reflection, this is not a good idea.

ppolewicz:
this reminds me of my first contribution: https://github.com/Backblaze/B2_Command_Line_Tool/pull/2#issuecomment-160005313

PR #337, opened 2017-04-11
Send file hashes at the end

svonohr:


b2/progress.py
```diff
@@ -228,3 +229,60 @@ def write(self, data):
     def _update(self, delta):
         self.bytes_completed += delta
         self.progress_listener.bytes_completed(self.bytes_completed + self.offset)
+
+
+class StreamWithHash(object):
+    """
+    Wraps a file-like object, caculates SHA1 while reading
```
ppolewicz: `s/cac/calc/`

b2/progress.py
```diff
@@ -228,3 +229,60 @@ def write(self, data):
...
+        self.hash_read = 0
+
+    def read(self, size=None):
+        data = b''
+        if self.hash is None:
+            # Read some bytes from stream
+            if size is None:
+                data = self.stream.read()
+            else:
+                data = self.stream.read(size)
```
ppolewicz: this line is not covered by UT

b2/bucket.py
```diff
@@ -387,15 +386,18 @@ def _upload_small_file(
...
                     progress_listener.set_total_bytes(content_length)
                     input_stream = StreamWithProgress(file, progress_listener)
-                    upload_response = self.api.raw_api.upload_file(
-                        upload_url, upload_auth_token, file_name, content_length, content_type,
-                        sha1_sum, file_info, input_stream
+                    hashing_stream = StreamWithHash(input_stream)
+                    length_with_hash = content_length + hashing_stream.hash_size()
+                    response = self.api.raw_api.upload_file(
+                        upload_url, upload_auth_token, file_name, length_with_hash, content_type,
+                        'hex_digits_at_end', file_info, hashing_stream
```
ppolewicz: `hex_digits_at_end` should be a constant

svonohr: Where should I put the constant? I also need it in raw_simulator.py

bwbeach: The `raw_simulator` already imports `raw_api`.  I'd put it in `raw_api`.

b2/bucket.py
```diff
@@ -510,9 +512,6 @@ def _upload_part(
 
         # Compute the SHA1 of the part
```
bwbeach: This comment is wrong now.

b2/progress.py
```diff
@@ -228,3 +229,60 @@ def write(self, data):
...
+
+            # Update hash
+            self.digest.update(data)
+
+            # Check for end of stream
+            if size is None or (len(data) == 0 and size != 0):
```
bwbeach: One more case to consider: size is non-None, but less data was read that requested, indicating EOF.

How about: `if size is None or len(data) < size:`

b2/progress.py
```diff
@@ -228,3 +229,60 @@ def write(self, data):
...
+                self.hash = self.digest.hexdigest()
+
+        if self.hash is not None:
+            # The end of stream was reached, return hash now
+            read_size = len(self.hash) if size is None else size
```
bwbeach: You may need to read less than `size` if you've already read some from the stream.

test/test_progress.py
```diff
@@ -0,0 +1,63 @@
...
+        self.assertEqual(self.expected, output)
+
+    def test_no_argument_more(self):
+        output = self.stream.read(len(self.data) + 1)
+        output += self.stream.read()
+        self.assertEqual(self.expected, output)
+
+    def test_one_by_one(self):
+        output = b''
+        data = self.stream.read(1)
```
bwbeach: It's probably worth checking that each read returns the right size.  I think doing so would catch the issue I mentioned above when deciding how much of the hash to return.

Maybe something like:

```
for expected_byte in self.expected:
    self.assertEqual(expected_byte, self.stream.read(1))
self.assertEqual(b'', self.stream.read(1))
```

bwbeach:
Nice work.  I like the way you've structured this.

b2/progress.py
svonohr: Is this really correct? I've already asked at Slack because the documentation at https://docs.python.org/3/library/io.html#io.RawIOBase.read suggests that reading 0 bytes indicates EOF and less bytes than requested may happen during normal operation.

bwbeach: The doc for file objects says:

Read at most size bytes from the file (less if the read hits EOF before obtaining size bytes). If the size argument is negative or omitted, read all data until EOF is reached. The bytes are returned as a string object. An empty string is returned when EOF is encountered immediately. (For certain files, like ttys, it makes sense to continue reading after an EOF is hit.) **Note that this method may call the underlying C function fread() more than once in an effort to acquire as close to size bytes as possible.** Also note that when in non-blocking mode, less data than was requested may be returned, even if no size parameter was given.

svonohr: Ok, I'll rewrite this tomorrow.

test/test_progress.py
svonohr: How do I write this for Python 2 and 3? With your version I get assertion failures that 48 is not equal to b'0' on Python 3. So I used `six.int2byte(expected_byte)` and now Python 2 complains that `expected_byte` is not an integer.

bwbeach:
Python 3 byte string subscripting is not consistent with strings.  For byte strings, `expected[0]` returns an int, while `expected[0:1]` returns a byte string.  After poking around in the docs, I found `six.iterbytes`, which behaves consistently in both:

```
    def test_one_by_one(self):
        for b in six.iterbytes(self.expected):
            self.assertEqual(six.int2byte(b), self.stream.read(1))
        self.assertEqual(b'', self.stream.read(1))
```

svonohr:
Thanks for fixing this.

bwbeach:
Thanks for implementing this feature!  I'll merge the pull request once the tests pass.

PR #343, opened 2017-05-18
Convert to Python3

ralic:
using : find . -name '*.py' | xargs 2to3 -w


svonohr:
B2 CLI is already compatible with Python 2 and 3, so what's the purpose of this?

ralic:
No need to make change for python3 compatible packages. Sorry.

PR #346, opened 2017-05-26
Add usage info about B2_ACCOUNT_INFO environment variable.

bwbeach:


b2/console_tool.py
```diff
@@ -1033,6 +1031,13 @@ def _usage_and_fail(self):
                 self._print_stderr(line)
 
         self._print_stderr('')
+        self._print_stderr(
+            'The environment variable B2_ACCOUNT_INFO specifies the sqlite file to use'
+        )
+        self._print_stderr(
+            'for caching authentication information.  Default is: ~/.b2_account_info'
```
ppolewicz: Now that we use it in two places, this should become a constant such as `SqliteAccountInfo.DEFAULT_DB_LOCATION`

PR #370, opened 2017-09-19
Add --failOnEmptySource to sync (fixes #369)

### LINKED ISSUE #369: Add option to exit sync if source directory is empty
An option to exit the `sync` command if the source directory is empty would be useful. Currently it's easy to run `sync` with a typo in the source path or against an unmounted source path and delete all files in a B2 bucket. Since the normal use case is to run against a non-empty source directory, an option to check if the source directory is non-empty and fail if it is would be a nice sanity check.

shane-axiom:
Adds `--failOnEmptySource` to the sync operation to prevent b2 sync from deleting all files in a bucket if an incorrect source path or path to an unmounted source volume is specified.


svonohr:
I haven't looked at the code in detail, but I think this should be the new default behavior. If someone really wants to do a sync with empty source path they should specify so.

shane-axiom:
The patch currently has `failOnEmptySource` default to false, so I'd need to update a few things including existing tests to change it to default to true.

bwbeach:
After thinking about it overnight, I agree with @svonohr that the default should be to *not* delete all your files.  :-)

It's true that it breaks compatibility, but I think that intentionally deleting all your files is rare enough that it OK.  And it will save some people a lot of grief.

b2/sync/sync.py
```diff
@@ -210,6 +218,9 @@ def sync_folders(
     if dest_folder.folder_type() == 'local' and not dry_run:
         dest_folder.ensure_present()
 
+    if source_folder.folder_type() == 'local' and not dry_run and fail_on_empty_source:
```
bwbeach: Any reason to skip the check on a dry run?

shane-axiom: Nope, I was just following the convention of the target directory check above. I'll remove the dry_run check.

ppolewicz:
While I agree that this should be the default behavior, I strongly recommend that this change is done with a top version increase (the first number, before the first dot).

Someone can have scripts on production using the old behavior. We also change the sync public API. In my opinion, a well maintained tool / library should not change the default behavior between middle or minor versions.

shane-axiom:
Perhaps the 0.x line gets `--failOnEmptySource` which defaults to false when the flag isn't provided, and the 1.x line changes to use `--allowEmptySource` (also defaulting to false when not provided)?

ppolewicz:
If it is not implemented in 0.x, everyone who will write their scripts on 0.x (using the old default behavior) will then have to modify their code for 1.x

On the other hand, there is no harm in implementing `--allowEmptySource` in 0.x and documenting it.

bwbeach:
In general, I agree with the principle that we should preserve compatibility within a major release number.

The argument on the other side is that people have lost data because it's too easy to delete everything.  I'll gather some input around the Backblaze office today.

bwbeach:
Also, before going to the next major version number, I would like to review the interface.  There are some inconsistencies, and it would be good to clean them up.

ppolewicz:
If people are loosing data because of this, we should release a version with the UI change quickly, but that doesn't mean that we should increase only a minor version.

bwbeach:
Let's update the major version number.

Go ahead, @shane-axiom, and update the option name, if you have time.  I'll take care of the version number when it's released.

I just reviewed the interface, and didn't see anything urgent that needs changing.

malcolmrh:
@bwbeach I changed the option to `--allowEmptySource` which defaults to False and tested, working for me.

shane-axiom:
@bwbeach I changed the option to --allowEmptySource which defaults to False and tested, working for me.

bwbeach:
Looks like there's a code formatting error.  The way to fix it is to install the latest yapf, and let the `pre-commit.sh` update the code, then commit the changes.

shane-axiom:
@bwbeach Ok, I ran pre-commit.sh and committed the formatting changes. With the latest version of yapf (0.18.0) there are a lot of formatting changes which touch parts of the code I didn't work on, is that desirable? Should I be using 0.17.0 instead?

bwbeach:
Thanks @shane-axiom!  

yapf is annoying because they keep changing their minds on formatting.  Including the files you didn't work on means that I won't have to run it myself.

Issue #334, opened 2017-04-05
Missing characters when syncing root


BobVul:
With an attempt at syncing root, e.g. `b2 sync / b2://foo/bar`, a character is missing in the path resulting in errors such as:

```
OSError: [Errno 2] No such file or directory: '/in/bash'
```

Which somehow found `/in/bash` instead of `/bin/bash`.

I believe this is because the attempt at calculating prefix length in [folder.py:79](https://github.com/Backblaze/B2_Command_Line_Tool/blob/0308d37e5cc92765064118b38abccd35351bfd11/b2/sync/folder.py#L79) fails to take into account the root path `/`, where there is no _trailing_ slash as such.

A quick fix might be to special-case root, but to be honest the whole prefix calculation looks pretty fragile.

ppolewicz:
I see that you have analysed it pretty well already. Would you take a shot at suggesting a fix for this issue?

We encourage outside contributors to perform changes on our codebase. Many such changes have been merged already. In order to make it easier to contribute, core developers of this project:

* provide guidance (through the issue reporting system)
* provide tool assisted code review (through the Pull Request system)
* maintain a set of integration tests (run with a production cloud)
* maintain a set of (well over a hundred) unit tests
* automatically run unit tests on 14 versions of python (including osx, Jython and pypy)
* format the code automatically using [yapf](https://github.com/google/yapf)
* use static code analysis to find subtle/potential issues with maintainability
* maintain other Continous Integration tools (coverage tracker)

Elusive138:
If we can make the assumption that this is only for slash-only paths (i.e. root and paths that collapse down to root), a basic fix might be to replace that line with:

```
prefix_len = len(self.root)
path_chars = set(self.root)
if len(path_chars) > 0 and path_chars != set('/'):
    prefix_len += 1 # when path contains more than just `/`, include trailing '/' in prefix length, see issue #334
```

But I'm not sure just how *safe* such a change would be, therefore I'm a bit hesitant with this.

For the time being, I think I'll be looking at rclone - I don't really have time to worry about B2 CLI reliability right now.


Just for completeness, here's the full stack trace

```
# b2 sync --dryRun / b2://foo/bar
ERROR:b2.console_tool:ConsoleTool unexpected exception
Traceback (most recent call last):
  File "/usr/local/lib/python2.7/dist-packages/b2/console_tool.py", line 992, in run_command
    return command.run(args)
[...]

PR #374, opened 2017-09-22
#334: Add prefix_len property to LocalFolder.

marchevskiy:
This PR attempts to resolve an edge case in the prefix length derivation of the LocalFolder class by using `os.path.splitdrive` to generalize cases when the root path is essentially both a leading and trailing delimiter. 

The unit tests are passing for me but I have not been able to test the windows compatibility. If the CI checks pass I will work on getting a VM setup.

### DIFF
--- a/b2/sync/folder.py
+++ b/b2/sync/folder.py
@@ -173,12 +173,20 @@ class LocalFolder(AbstractFolder):
         # Check whether a path is a relative symlink.
         #
         def validate(x):
+            # type: (str) -> bool
             if not platform.system() == 'Linux':
                 return True  # TODO: Not sure about how this should work in Windows.
-            try:
-                str(os.readlink(x)) not in ['.', '..']
-            except OSError as e:
-                return e.errno == errno.EINVAL
+            if os.path.islink(x):
+                try:
+                    # Check if symlink points up in the directory tree.
+                    a = not os.readlink(x).startswith('../')
+                    # Check if symlink is recursive.
+                    b = os.readlink(x) != '.'
+                    return not(a or b)
+                except OSError as e:
+                    return e.errno == errno.EINVAL
+            else:
+                return True
 
         # Yield all of the answers
         for name in sorted(names):
@@ -187,6 +195,8 @@ class LocalFolder(AbstractFolder):
                 if validate(full_path):  # Don't iterate over relative symlinks.
                     for rp in self._walk_relative_paths(prefix_len, full_path, reporter):
                         yield rp
+                else:
+                    yield relative_path
             else:
                 yield relative_path
 
--- a/b2/sync/folder.py
+++ b/b2/sync/folder.py
@@ -71,8 +71,7 @@ class LocalFolder(AbstractFolder):
         if not isinstance(root, six.text_type):
             raise ValueError('folder path should be unicode: %s' % repr(root))
 
-        # Using normpath here allows windows and linux paths to be handled.
-        self.root = os.path.normpath(os.path.abspath(root))
+        self.root = os.path.abspath(root)
 
     @property
     def prefix_len(self):
@@ -81,8 +80,7 @@ class LocalFolder(AbstractFolder):
             Prefix length calculation:
                 https://github.com/Backblaze/B2_Command_Line_Tool/issues/334
 
-            Using os.path.normpath and os.path.abspath, there are three cases to
-                handle here:
+            Using os.path.abspath, there are three cases to handle here:
                   * root paths with a drive specification (windows),
                   * root paths without a drive specification (not windows),
                   * non root paths
@@ -94,22 +92,25 @@ class LocalFolder(AbstractFolder):
                 >>> import ntpath
                 >>> posix_root = '/'
                 >>> nt_root = 'C://'
-                >>> posixpath.splitdrive(posixpath.normpath(posix_root))
+                >>> posixpath.splitdrive(posixpath.abspath(posix_root))
                 ('', '/')
                 >>> map(len, _)
                 [0, 1]
-                >>> ntpath.splitdrive(ntpath.normpath(nt_root))
+                >>> ntpath.splitdrive(ntpath.abspath(nt_root))
                 ('C:', '\\')
                 >>> map(len, _)
                 [2, 1]
 
+            Furthermore, there is a trailing separator only if the path denotes a root filesystem.
+
         """
 
         drive, tail = os.path.splitdrive(self.root)
-        if len(tail) == 1:
+        if len(tail) == 1:  # If the tail is only one character it must be / or \\
+            # Include the drive if necessary and treat the leading separator as a trailing one.
             return len(drive) + len(tail)
         else:
-            return len(self.root) + 1
+            return len(self.root) + 1  # include trailing '/' in prefix length
 
     def folder_type(self):
         return 'local'
--- a/b2/sync/folder.py
+++ b/b2/sync/folder.py
@@ -70,14 +70,52 @@ class LocalFolder(AbstractFolder):
         """
         if not isinstance(root, six.text_type):
             raise ValueError('folder path should be unicode: %s' % repr(root))
-        self.root = os.path.abspath(root)
+
+        # Using normpath here allows windows and linux paths to be handled.
+        self.root = os.path.normpath(os.path.abspath(root))
+
+    @property
+    def prefix_len(self):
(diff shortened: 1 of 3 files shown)


b2/sync/folder.py
```diff
@@ -70,14 +70,52 @@ def __init__(self, root):
...
+        if len(tail) == 1:
+            return len(drive) + len(tail)
+        else:
+            return len(self.root) + 1
 
     def folder_type(self):
         return 'local'
 
     def all_files(self, reporter):
-        prefix_len = len(self.root) + 1  # include trailing '/' in prefix length
```
bwbeach: We've lost the comment saying that the trailing "/" is in the prefix length.  I think that would be worth saying somewhere.

It's probably also worth saying that self.root has a trailing "/" only if it's the root of the filesystem.

test/test_sync.py
```diff
@@ -105,6 +107,58 @@ def test_invalid_permissions(self):
...
+            # Build a folder.
+            folder = LocalFolder(base_path)
+
+            # Verify that the prefix truncates the root path correctly.
+            self.assertEqual(folder.root[folder.prefix_len:], u'')
+
+            # Verify full path to relative path conversion.
+            test_path = folder.root
+            test_sub_path = u'test'
+            if not folder.root.endswith(separator):
```
bwbeach: This "if" is what os.path.join does:

```
>>> os.path.join('/foo/', 'bar')
'/foo/bar'
>>> os.path.join('/foo', 'bar')
'/foo/bar'
```

test/test_sync.py
```diff
@@ -105,6 +107,58 @@ def test_invalid_permissions(self):
                 os.path.join(tmpdir, self.NAMES[0])
             )
 
+    def test_prefix_calculation(self):
+        """
```
bwbeach: I like the comments.  Thorough and easy to read.

test/test_sync.py
```diff
@@ -105,6 +107,58 @@ def test_invalid_permissions(self):
...
+
+            For windows, case 1) is different (e.g. C:\\) but the others should be the same.
+
+        """
+
+        prefix = os.path.normpath(u'C://') if platform.system() == 'Windows' else unicode(os.path.sep)
+        separator = unicode(os.path.sep)
+        root = os.path.normpath(os.getcwdu())  # Always has leading but not trailing separator.
+        relative = u''.join(choice(lowercase) for _ in range(randint(4, 8)))
+        permutations = dict(**{
```
bwbeach: How about just `{ ... }` instead of `dict(**{ ... })`?

test/test_sync.py
marchevskiy: Yep good catch.

test/test_sync.py
marchevskiy: Yep sorry, got fooled by my IDE there.

b2/sync/folder.py
marchevskiy: I'll add it back in. There's somewhat of an explanation within the docstring for the `prefix_len` property, but it is not as explicit. 

Thinking about this a bit more, I believe there's no reason to keep passing `prefix_len` to `_walk_relative_paths` since it's a property of the class now.

b2/sync/folder.py
```diff
@@ -70,14 +70,52 @@ def __init__(self, root):
         """
         if not isinstance(root, six.text_type):
             raise ValueError('folder path should be unicode: %s' % repr(root))
-        self.root = os.path.abspath(root)
+
+        # Using normpath here allows windows and linux paths to be handled.
+        self.root = os.path.normpath(os.path.abspath(root))
```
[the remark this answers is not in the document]
marchevskiy: Yeah I was a bit confused by the documentation here, and think I got it backwards. It's definitely a no-op here and I'll remove it.

marchevskiy:
The Travis CI checks appear to be failing because string.lowercase doesn't exist for python3. I'll find a way to use the six module or just generate a random string differently.

svonohr:
By chance I thought about this bug already months ago and came up with a different solution:
`prefix_len = len(os.path.join(self.root, ''))`
By appending an empty string the join will add a trailing slash to the path, if one does not already exist, or if its the root of the filesystem it also won't add another slash. I'm not sure however, if this covers all cases this PR covers or if this behavior of `join` is standard.

marchevskiy:
That's certainly a more elegant solution. I'm new to the github ecosystem so I'm not sure if you can just commit to my branch?

svonohr:
You'd have to add me as a collaborator to your fork. Or I could fork your branch, do the changes in my branch and then create a new pull request. It's probably easier if you just change it in your branch (and I don't have much time at the moment).

marchevskiy:
Ok just wanted to make sure you get credit. I'll update this branch and hopefully it will be ready to merge.

svonohr:
I haven't tested your branch yet, but I believe there is another place which doesn't handle the root path properly. In sync/folder_parser.py line 30 we have the following code:
```
if folder_name.endswith('/'):
    folder_name = folder_name[:-1]
return LocalFolder(folder_name)
```
This will just remove the whole string when it's just the root path.

marchevskiy:
Yeah it will end up using the current working directory instead of the path in `folder_name`. The fix is trivial, just let `LocalFolder` take care of the root prefix and don't pre-process in higher level functions. 

I think it might be worthwhile to write some integration unit tests that mock out os.path* to get better coverage in this area. I can take a shot at drafting something up for that in a different branch.

marchevskiy:
I stumbled on an issue when testing this in my VM where some files had relative symlinks - e.g. `/usr/bin/X11 -> .` - and would cause an infinite recursion loop in `_walk_relative_paths` so I added a bit of protective code against that case.

svonohr:
Isn't that a general problem with cyclic symlinks? I tried synching such a cyclic directory and b2 cli builds a path of arround 500 characters and then complaints that the symlink seems to be broken.

marchevskiy:
Yeah the validate closure had some issues. Can you please provide an example of the cyclical directory structure you were able to produce a broken symlink with?

svonohr:
You don't even have to use relative links. I've created some directories `/foo/bar` and then added an absolute link to `/foo`. This way b2 cli will come up with paths like `/foo/bar/link/bar/link/...`. I'm not sure how one would prevent this.

ppolewicz:
[os.path.islink](https://docs.python.org/3/library/os.path.html#os.path.islink) and [os.path.realpath](https://docs.python.org/3/library/os.path.html#os.path.realpath), maybe [os.path.commonprefix](https://docs.python.org/3/library/os.path.html#os.path.commonprefix)

marchevskiy:
@svonohr I believe that will not occur on this branch, perhaps you were on the master branch?

Master:
``` 
root@2db87b17b227:/# mkdir -p /foo/bar/
root@2db87b17b227:/# ln -s /foo /foo/bar/link
root@2db87b17b227:/# b2 sync --dryRun /foo/ b2://bucket-74358734
WARNING: /foo/bar/link/bar/link/bar/link/bar/link/bar/link/bar/link/bar/link/bar/link/bar/link/bar/link/bar/link/bar/link/bar/link/bar/link/bar/link/bar/link/bar/link/bar/link/bar/link/bar/link/bar/link/bar/link/bar/link/bar/link/bar/link/bar/link/bar/link/bar/link/bar/link/bar/link/bar/link/bar/link/bar/link/bar/link/bar/link/bar/link/bar/link/bar/link/bar/link/bar/link/bar/link could not be accessed (broken symlink?)
```

issues/334:
```
root@2db87b17b227:/# mkdir -p /foo/bar/
root@2db87b17b227:/# ln -s /foo /foo/bar/link
root@2db87b17b227:/# b2 sync --dryRun /foo/ b2://bucket-74358734
upload bar/link      
```

...though if I do actually perform the sync, I'll get an exception thrown for an invalid source: 

`b2_upload(/foo/bar/link, bar/link, 1507226301968): InvalidUploadSource('/foo/bar/link',) Invalid upload source: /foo/bar/link`. 

I'm not *exactly* sure how symlinks are intended to be handled, so I could use some help from the developers in understanding what the expected behavior is.

bwbeach:
Thanks, @marchevskiy, for the all the work you did on this.  Your work made it clear that my initial approach on this code was just wrong.  I tried having the code keep two paths: the path on the local disk, and the B2 path.  That made things a lot clearer.  See #384.

PR #375, opened 2017-09-27
Add --compareThreshold and --verbose

shane-axiom:
Add `--compareThreshold` to allow deltas in file modTime or size comparisons. Also add `--verbose` for simple verbose (debug) logs to stdout.

### DIFF
--- a/b2/console_tool.py
+++ b/b2/console_tool.py
@@ -83,7 +83,7 @@ class Command(object):
     OPTION_FLAGS = []
 
     # Global option flags.  Not shown in help.
-    GLOBAL_OPTION_FLAGS = ['debugLogs']
+    GLOBAL_OPTION_FLAGS = ['debugLogs', 'verbose']
 
     # Explicit arguments.  These always come before the positional arguments.
     # Putting "color" here means you can put something like "--color blue" on
@@ -684,9 +684,10 @@ class MakeUrl(Command):
 class Sync(Command):
     """
     b2 sync [--delete] [--keepDays N] [--skipNewer] [--replaceNewer] \\
-            [--compareVersions <option>] [--threads N] [--noProgress] \\
-            [--excludeRegex <regex> [--includeRegex <regex>]] [--dryRun] \\
-            [--allowEmptySource] <source> <destination>
+            [--compareVersions <option>] [--compareThreshold N] \\
+            [--threads N] [--noProgress] [--dryRun ] [--allowEmptySource ] \\
+            [--excludeRegex <regex> [--includeRegex <regex>]] \\
+            <source> <destination>
 
         Copies multiple files from source to destination.  Optionally
         deletes or hides destination files that the source does not have.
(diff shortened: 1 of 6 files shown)


b2/console_tool.py
```diff
@@ -739,6 +740,12 @@ class Sync(Command):
         A future enhancement may add the ability to compare the SHA1 checksum
         of the files.
 
+        Fuzzy comparison of files based on modTime or size can be enabled by
+        specifying the --compareThreshold option. This will allow a difference
```
bwbeach: (Sorry for the delay in commenting on this.  This is a vacation week for me.)

Nice work on the coding, documenting, and testing.

My only suggestion is to clarify this documentation a little bit.  Instead of "allow a difference", 
I might say something like:  

```
This will create modTimes (in milliseconds) or sizes (in bytes) as the same if they are within the comparison threshold.  Files that match, within the threshold, will not be synced.
```

shane-axiom: @bwbeach Thanks, I revised the documentation text based on your suggestion. The `--compareThreshold` section now reads:

>  Fuzzy comparison of files based on modTime or size can be enabled by
specifying the --compareThreshold option.  This will treat modTimes
(in milliseconds) or sizes (in bytes) as the same if they are within
the comparison threshold.  Files that match, within the threshold, will
not be synced. Specifying --verbose and --dryRun can be useful to
determine comparison value differences.

b2/sync/policy.py
```diff
@@ -67,30 +75,45 @@ def files_are_different(cls, source_file, dest_file, args):
...
+            logger.debug(
+                'File %s: source time %s, dest time %s, diff %s, threshold %s, diff > threshold %s',
+                source_file.name, source_mod_time, dest_mod_time, diff_mod_time, compareThreshold,
+                diff_mod_time > compareThreshold
+            )
 
-            # Source is older
-            elif source_mod_time < dest_mod_time:
-                if args.replaceNewer:
+            if diff_mod_time > compareThreshold:
```
ppolewicz: this is done twice - here and in the log above. If it is done in one place and assigned to a variable, there will be no risk of someone changing one of those and forgetting about the other one.

b2/sync/sync.py
```diff
@@ -260,7 +260,7 @@ def sync_folders(
         for action in make_folder_sync_actions(
             source_folder, dest_folder, args, now_millis, reporter
         ):
-            logging.debug('scheduling action %s on bucket %s')
+            logging.debug('scheduling action %s on bucket %s', action, bucket.name)
```
ppolewicz: `Bucket` has a very good `__repr__`, maybe use that?

shane-axiom:
@ppolewicz PR is updated.

b2/console_tool.py
```diff
@@ -1008,9 +1016,13 @@ def run_command(self, argv):
             logger.info('ConsoleTool \'args is None\' - printing usage')
             self._print_stderr(command.command_usage())
             return 1
-        elif args.logConfig and args.debugLogs:
-            logger.info('ConsoleTool \'args.logConfig and args.debugLogs\' were both specified')
-            self._print_stderr('ERROR: --logConfig and --debugLogs cannot be used at the same time')
+        elif [args.logConfig, args.verbose, args.debugLogs].count(True) > 1:
```
ppolewicz: You should use a tuple here, which is faster to initialize and doesn't need to allocate memory ahead to amortize `,append()` calls.

`.count()` usage here is very clever!

bwbeach: Pawel is correct that tuples are faster, but this only happens once per invocation of the command-line tool, so I'm not worried about it.  I'm going to go ahead and merge the pull request.

shane-axiom: @ppolewicz Thanks, python isn't my main language so advice like this is helpful. I'll use a tuple next time!

PR #384, opened 2017-11-02
Fix #334 

### LINKED ISSUE #334: Missing characters when syncing root

bwbeach:


bwbeach:
Just tested syncing `C:/` on windows.

PR #385, opened 2017-11-02
Fixed off by one issue

### LINKED ISSUE #377: Off by one error in raw_api._download_file_from_url
I keep thinking this must be an error on my part but I can't see how I'm wrong.

I'm calling the following function:

```python
def download_file_by_id(
        self, download_url, account_auth_token_or_none, file_id, download_dest, range_=None
    )
```

The [documentation](https://www.backblaze.com/b2/docs/b2_download_file_by_id.html) for downloading a file says about the range parameter:

> The value "bytes=0-99" selects bytes 0 through 99 (inclusive) of the file, so it will return the first 100 bytes

range_ is specified as a tuple (in _download_file_from_url)

```python
if range_ is not None:
    assert len(range_) == 2, range_
    assert (range_[0] + 0) <= (range_[1] + 0), range_  # not strings
    assert range_[0] >= 0, range_
    assert range_[1] >= 1, range_
    request_headers['Range'] = "bytes=%d-%d" % range_
```
So When I want the first 1048576 bytes and I request the range `(0, 1048576 - 1)`. The response header contains `'Content-Length', '1048576'` which is correct. But later on there is the following validation:

```python
desired_length = range_[1] - range_[0]
if bytes_read != desired_length:
    raise TruncatedOutput(bytes_read, desired_length)
```

But that fails with the message:
> TruncatedOutput: only 1048576 of 1048575 bytes read

I can only come to the conclusion that the desired length should have been:

```python
desired_length = range_[1] - range_[0] + 1
```

But I'm confused because I wouldn't expect me to be the only one to have this issue.

obeleh:
Fixes #377

I don't think a test is really needed since it's checked by this line:
```python
if bytes_read != desired_length:
    raise TruncatedOutput(bytes_read, desired_length)
```

### DIFF
--- a/b2/raw_api.py
+++ b/b2/raw_api.py
@@ -277,7 +277,7 @@ class B2RawApi(AbstractRawApi):
                             actual=digest.hexdigest()
                         )
                 else:
-                    desired_length = range_[1] - range_[0]
+                    desired_length = range_[1] - range_[0] + 1
                     if bytes_read != desired_length:
                         raise TruncatedOutput(bytes_read, desired_length)
 


bwbeach:
Looks like there's a new version of YAPF (the code formatting tool), so the build thinks the formatting of existing code is wrong.  I'll make a pull request with the code formatted by the latest version.

The branch is `yapf-0.19.0` if you want to merge it in and fix the build.

bwbeach:
Making an integration test for this is hard without a way to request a range on the command line.  We can let that slide for now.  

The formatting issue is fixed in master, and won't conflict with this change, so I'll go ahead and merge.

Issue #397, opened 2017-12-09
sync errors if file is inaccessible after initial scan


shane-axiom:
sync can run into errors on huge directories if a file is moved, deleted, or becomes otherwise inaccessible after the intial scan of files to be synced is built.

Fix in PR incoming, this is just a placeholder issue.

Issue #265, opened 2016-10-23
b2 sync on Windows fails on long paths


ncote:
Several of the following warnings appear:
`WARNING:  <long path e.g. 279 characters in one case> could not be accessed (broken symlink?)`

Then it finally fails with:

```
File "C:\Python27\lib\site-packages\b2\sync.py", line 540, in _walk_relative_paths
  for name in os.listdir(dir_path):
WindowsError: [Error 3] The system cannot find the path specified: u'<257-character-long-path>\\*.*'
```

ppolewicz:
May be https://bugs.python.org/issue18199

ncote:
That looks right.  Should I assume then that the issue is unfixable here until the python bugs are fixed?

ppolewicz:
No. I am fairly certain that this can be fixed.

PR #398, opened 2017-12-09
Fix Windows integration tests, long Windows paths, and stale sync files

shane-axiom:
Mainly this is a fix for https://github.com/Backblaze/B2_Command_Line_Tool/issues/265 and https://github.com/Backblaze/B2_Command_Line_Tool/issues/397, with some other bonuses thrown in.  Comments welcome.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -201,10 +201,14 @@ before_install:
         fi
   - virtualenv --version
   - python --version
+  - |
+        # upgrade setuptools if it's too old to support PEP 508
+        # PEP 508 support is needed for python version dependent dependencies in requirements.txt/setup.py (e.g. futures)
+        pip install "setuptools>=20.2"
 
 install:
   # installing regex on jython fails (jython can't compile the native extension), this awk filter prevents it.
-  - cat requirements*.txt | awk \"${JYTHON:-False}'" == "True" && /^regex/ || ! /^regex/' | xargs pip install
+  - cat requirements*.txt | awk \"${JYTHON:-False}'" == "True" && /^regex/ || ! /^regex/' | cut -d \; -f 1 | xargs pip install
 
   - |
         # coverage has additional requirements
--- a/b2/sync/folder.py
+++ b/b2/sync/folder.py
@@ -8,15 +8,15 @@
 #
 ######################################################################
 
-from abc import ABCMeta, abstractmethod
 import os
+import six
 import sys
 
-import six
-
+from abc import ABCMeta, abstractmethod
 from .exception import EnvironmentEncodingError
 from .file import File, FileVersion
 from ..raw_api import SRC_LAST_MODIFIED_MILLIS
+from ..utils import is_file_readable
 
 
 @six.add_metaclass(ABCMeta)
@@ -155,13 +155,7 @@ class LocalFolder(AbstractFolder):
             local_path = os.path.join(local_dir, name)
             b2_path = join_b2_path(b2_dir, name)
             # Skip broken symlinks or other inaccessible files
-            if not os.path.exists(local_path):
-                if reporter is not None:
-                    reporter.local_access_error(local_path)
-            elif not os.access(local_path, os.R_OK):
-                if reporter is not None:
-                    reporter.local_permission_error(local_path)
-            else:
+            if is_file_readable(local_path, reporter):
                 if os.path.isdir(local_path):
                     name += six.u('/')
                 names.append((name, local_path, b2_path))
@@ -175,10 +169,13 @@ class LocalFolder(AbstractFolder):
                 for subdir_file in self._walk_relative_paths(local_path, b2_path, reporter):
                     yield subdir_file
             else:
-                file_mod_time = int(round(os.path.getmtime(local_path) * 1000))
-                file_size = os.path.getsize(local_path)
-                version = FileVersion(local_path, b2_path, file_mod_time, 'upload', file_size)
-                yield File(b2_path, [version])
+                # Check that the file still exists and is accessible, since it can take a long time
+                # to iterate through large folders
+                if is_file_readable(local_path, reporter):
+                    file_mod_time = int(round(os.path.getmtime(local_path) * 1000))
+                    file_size = os.path.getsize(local_path)
+                    version = FileVersion(local_path, b2_path, file_mod_time, 'upload', file_size)
+                    yield File(b2_path, [version])
 
     def _handle_non_unicode_file_name(self, name):
         """
--- a/b2/sync/folder.py
+++ b/b2/sync/folder.py
@@ -16,7 +16,7 @@ from abc import ABCMeta, abstractmethod
 from .exception import EnvironmentEncodingError
 from .file import File, FileVersion
 from ..raw_api import SRC_LAST_MODIFIED_MILLIS
-from ..utils import is_file_readable
+from ..utils import fix_windows_path_limit, is_file_readable
 
 
 @six.add_metaclass(ABCMeta)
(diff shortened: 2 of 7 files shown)


shane-axiom:
Looks like the version of setuptools used for Python 3.3 in the travis tests might be ancient, I'll look into asserting a minimum version to support PEP 508.

shane-axiom:
Added code in .travis.yml to upgrade setuptools if it's ancient (for Python 3.3) and some code in setup.py to assure a minimal setuptools version (21.0) for PEP 508 support. This is ready for review.

.travis.yml
```diff
@@ -171,10 +171,16 @@ before_install:
         fi
   - virtualenv --version
   - python --version
+  - |
+        # upgrade setuptools if it's too old to support PEP 508
+        if [ $(pip list 2>/dev/null | grep setuptools | grep -oP '\d+' | head -n 1) -lt 21 ]; then
```
ppolewicz: Instead of doing this (seems rather complex), could you just check if setuptools version is >= 20.2? Also, please explain in the comment why do we need setuptools to support PEP-508. This way, when one day we no longer need that due to refactor or something, we'll know that we can remove it. (well, we can guess, because between now and then another reason to have it in this version may appear ;) )

In the future, could you please surround `$(...)` with double quotes `"$(...)"? In this project we have a policy to quote all $expressions. Probably should be mentioned somewhere in developer info in `README.md`...

b2/sync/folder.py
```diff
@@ -113,6 +114,20 @@ def ensure_non_empty(self):
         if not os.listdir(self.root):
             raise Exception('Directory %s is empty' % (self.root,))
 
+    def _check_file_access(self, local_path, reporter):
```
ppolewicz: `_is_file_accessible` would be a better name. You wouldn't need a docstring then.

As far as I know, the only safe way to safely handle file access checking, is to try to open it to actually do stuff and catch the exception when that happens. Let me show you an example:

Thread-1: _check_file_access -> True
Thread-2: change file permissions, or move the file, or remove the file, unmount the filesystem etc
Thread-1: attempt to access the file and fail ("unexpectedly")

Now, I think the main purpose of this method is to attempt to report read access to the reporter, as well as it is possible... But the comment or the docstring of it doesn't make it clear. Since you are refactoring it to an own method, could you please mention this? That way, when one day someone will want to use this method, he won't be misled into thinking that it actually solves the problem.

setup.py
```diff
@@ -17,11 +17,16 @@
...
 # Always prefer setuptools over distutils
-from setuptools import setup, find_packages
+from setuptools import setup, find_packages, __version__ as setuptoolsversion
 # To use a consistent encoding
 from codecs import open
 from os import path
 
+#require at least setuptools 21.0 for PEP 508 conditional dependency support
+MIN_SETUPTOOLS_VERSION = (21, 0)
+if tuple(int(x) for x in setuptoolsversion.split('.')[:2]) < MIN_SETUPTOOLS_VERSION:
```
ppolewicz: split has a [maxsplit](https://docs.python.org/2/library/stdtypes.html#str.split) argument you can use instead of indexing here

requirements.txt
```diff
@@ -3,3 +3,4 @@ logfury>=0.1.2
 requests>=2.9.1
 six>=1.10
 tqdm>=4.5.0
+futures>=3.0.5; python_version < '3.2'
```
ppolewicz: why is this better?

setup.py
shane-axiom: @ppolewicz Unless I'm using it wrong, using maxsplit leaves the minor, patch, and any sub-patch versions unsplit.

```
>>> '21.1.0'.split('.',1)
['21', '1.0']
```

My goal here is to get the major and minor versions into a tuple for easy comparison, so it seems like I do need to do the slicing?

requirements.txt
shane-axiom: @ppolewicz I don't feel very strongly about this, but I tried to install dependencies using `pip install -r requirements.txt` under Python 2.7 and the `futures` dependency wasn't installed, since it's only specified in setup.py. PEP 508 seems to be the official recommended way to specify version dependent dependencies.

.travis.yml
```diff
@@ -171,10 +171,15 @@ before_install:
         fi
   - virtualenv --version
   - python --version
+  - |
+        # upgrade setuptools if it's too old to support PEP 508
+        # PEP 508 support is needed for python version dependent dependencies in requirements.txt/setup.py (e.g. futures)
+        pip install "setuptools>=20.2"
```
shane-axiom: @ppolewicz I think you mean like this? Yes, much simpler!

setup.py
ppolewicz: oops, you are correct

setup.py
```diff
@@ -17,11 +17,16 @@
...
-from setuptools import setup, find_packages
+from setuptools import setup, find_packages, __version__ as setuptoolsversion
 # To use a consistent encoding
 from codecs import open
 from os import path
 
+#require at least setuptools 20.2 for PEP 508 conditional dependency support
+MIN_SETUPTOOLS_VERSION = (20, 2)
+if tuple(int(x) for x in setuptoolsversion.split('.')[:2]) < MIN_SETUPTOOLS_VERSION:
+    sys.exit("setuptools %s.%s or later is required." % MIN_SETUPTOOLS_VERSION)
```
ppolewicz: some people will see this and create an issue in our github issue tracker asking how to resolve this issue

I suggest you add ` To fix it, try running this: pip install "setuptools>=20.2"`.

shane-axiom:
@ppolewicz I think I've addressed (or attempted to address) all of your comments/concerns. This should be ready for another round of review or merging. Thanks!

.travis.yml
```diff
@@ -171,10 +171,15 @@ before_install:
...
+  - pip list
```
ppolewicz: we already have `pip freeze` in `before_script`. We don't really install anything until this point (except maybe setuptools), I can't see what good does putting pip list here give us. Could you please help me understand it (or remove the line)?

b2/sync/folder.py
```diff
@@ -113,6 +114,20 @@ def ensure_non_empty(self):
         if not os.listdir(self.root):
             raise Exception('Directory %s is empty' % (self.root,))
 
+    def _is_file_readable(self, local_path, reporter):
+        """
+        Check if a file exists and is readable
```
ppolewicz: I'd remove the docstring completely, because the function is trivial to understand and one day someone will add a third `if` statement down below and forget to update the docstring.

b2/sync/folder.py
```diff
@@ -113,6 +114,20 @@ def ensure_non_empty(self):
         if not os.listdir(self.root):
             raise Exception('Directory %s is empty' % (self.root,))
 
+    def _is_file_readable(self, local_path, reporter):
```
ppolewicz: I suggest `reporter=None` as the method is clearly expecting this, this way we can let the user of this method do `self._is_file_readable(local_path)` instead of `self._is_file_readable(local_path, reporter=None)` which is a bit tedious or `self._is_file_readable(local_path, None)` which looks weird.

I am also thinking if `_is_file_readable` should be located in `sync.folder.AbstractFolder()` class? It strikes me that a subclass is unlikely to shadow it and the method is not referencing `self` at any point. I think a more appropriate place could be a function in `b2.utils`. Another argument is that for testing this method we don't really need an instance of AbstractFolder subclass (that kind of test would look rather awkward, we'd have to write a class that inherits from AbstractFolder... to work around that we could make `_is_file_readable` a classmethod, but that's still clearly not perfect, I think b2.utils is the way to go). If you agree, please move it, alternatively lets discuss it.

shane-axiom:
@ppolewicz Ok, I applied your last round of suggestions. Not sure why the Jython Travis build is failing now, I don't think anything I touched should have affected it. Other than that, this should be ready.

bwbeach:
I just fixed the jython build and merged the fix into master.  If you merge from there it should clear up the build.

Issue #364, opened 2017-08-23
Directories excluded by excludeRegex can still give permission errors


jwiegley:
I have the following sync command:

```
backblaze-b2 sync --keepDays 722 --replaceNewer \
    --excludeRegex '(.*\.DS_Store)|(.*\.Spotlight-V100.*)|(.*\.Trashes.*)|(.*\.fseventsd.*)|(.*\.DocumentRevisions-V100.*)|(.*\.TemporaryItems)' \
    /Volumes/mybook/Audio b2://XXX
```

However, despite Trashes being ignored, I still get:

```
WARNING: /Volumes/mybook/Audio/.Trashes could not be accessed (no permissions to read?)
```

ppolewicz:
That's a bug. The way `--excludeRegex` is implemented, it first makes a list of everything (including things that sync shouldn't touch) and then it filters items. This can cause a directory with a very large number of files to be listed unnecessarily, wasting time, wasting CPU and potentially causing warnings as in your case.

dmanikowski-reef:
Hi guys, I looked into it. It is quite easy to replace the list @ppolewicz is talking about with the generator but there is a minor issue. List of files to be copied is prepared using all_files() method of the LocalFolder class. As far as I know this method should return SORTED list of all files in the directory which we are unable to achieve using generator. We can sort the list afterwards if needed but I don't know if this would be expected behavior. Is this sorting really needed in all_files() method?

ppolewicz:
I believe sync uses the sorting later on. We can move the scanning to another (sub)method which returns a generator, filter right there and then cast to list and perform the sorting in all_files()

bwbeach:
The doc for `sync` says:

    You can specify --excludeRegex to selectively ignore files that
    match the given pattern. Ignored files will not copy during
    the sync operation. The pattern is a regular expression
    that is tested against the full path of each file.

I think that excluding `.*\.DS_Store` should not exclude `.DS_Store/a.txt`, because the regex doesn't match the full file name.  To exclude it, you would need a regex like `.*\.DS_Store/.*`.

bwbeach:
#407 took care of this one.

marcobarriosfi:
This problem still exists.

ppolewicz:
@marcobarriosfi which version of the CLI are you using?

marcobarriosfi:
Seems to be 1.1.0-1. Installed it via apt.

ppolewicz:
[...]

PR #407, opened 2018-02-11
Fix filtering on sync

krzysztofgromadzki:
Issue: https://github.com/Backblaze/B2_Command_Line_Tool/issues/364

With this implementation we avoid making a list of a whole
directory if we want to exclude some files/dirs.
API of the all_files method changed a bit but it was necessary.
Tests needed to be modified as well due to API changes.


ppolewicz:
Build is failing due to https://github.com/crsmithdev/arrow/issues/517

lets wait a few days, perhaps they will fix it there

bwbeach:
Thanks for working on this.  Sorry I'm late to the party.

The doc for `sync` says:

    You can specify --excludeRegex to selectively ignore files that
    match the given pattern. Ignored files will not copy during
    the sync operation. The pattern is a regular expression
    that is tested against the full path of each file.

So I don't think it's correct to match partial file names.  The exclusion regex `dirName` should not match the file `dirName/file.txt`.

I could image teaching the code that `dirName/.*` can exclude the directory.  But excluding the entire directory is not correct if there is an inclusion regex like `.*\.txt` that could match files inside that directory.

ppolewicz:
Would the current optimization be correct if there is no `--includeRegex` specified?

bwbeach:
Based on the current code, and my interpretation of the doc, what's there is not quite correct. Specifying `--excludeRegex 'dirName'` should not exclude the file `dirName/file.txt`.

Adding a special case for `.*` at the end of the regex, and saying that directories are only excluded if there is no `--includeRegex` would work.  It would just rule out some use cases.

This area always gets annoyingly complicated.

Another option would be to add a new option `--excludeDirRegex` that is defined to exclude everything in a directory, even if it matches an inclusion.

ppolewicz:
I think `--excludeDirRegex` is a good idea.

bwbeach:
The scanning methods, such as `_walk_relative_paths` already take several parameters, and I'm reluctant to add another one for excluding directories.  How about defining a `FileFilter` interface, and  replacing the `exclusions` and `inclusions` with a `FileFilter`?  It could handle filtering both files and directories:

```
@six.add_metaclass(ABCMeta)
class FileFilter(object):
    """
    Policy object that decides which files should be included in a scan.
    """

    @abstractmethod
    def should_exclude_file(self, file_path):
        """
        Given the full path of a file, should it be excluded from the scan?

        :param file_path: The path of the file, relative to the root directory
                          being scanned.  This is a file, not a directory, so
                          the path will never end in '/'.
        :return: True iff excluded.
        """

    @abstractmethod
    def should_exclude_directory(self, dir_path):
        """
        Given the full path of a directory, should all of the files in it be
        excluded from the scan?

        :param dir_path: The path of the directory, relative to the root directory
                         being scanned.  This is a directory, so the path will
                         always end in '/'.
        :return: True iff excluded .
        """
```

b2/sync/filters.py
```diff
@@ -0,0 +1,129 @@
+######################################################################
+#
+# File: b2/sync/filters.py
+#
+# Copyright 2016 Backblaze Inc. All Rights Reserved.
```
bwbeach: 2018

b2/sync/filters.py
```diff
@@ -0,0 +1,129 @@
...
+import os.path
+from abc import ABCMeta, abstractmethod
+
+logger = logging.getLogger(__name__)
+
+
+@six.add_metaclass(ABCMeta)
+class FileFilter(object):
+    @abstractmethod
+    def is_exclude_filter(self):
```
bwbeach: An interface that lets you ask what type of object you have usually means that the abstraction is could be better.

A class docstring saying what the class does might help.  From a quick look at the code, I think that: (1) there are filters and a filter manager, (2) the filter manager is what combines filters, and (3) the filter manager is what client code will use when testing whether a file or directory should be included.  Is that right?

b2/sync/filters.py
```diff
@@ -0,0 +1,129 @@
...
+
+class FilterManager(object):
+
+    FILTERS_PRIORITES = {
+        ExcludeDirRegexFilter: 100,
+        IncludeFileRegexFilter: 50,
+        ExcludeFileRegexFilter: 0,
+    }
+
+    def __init__(self, filters):
```
bwbeach: I think that having the three filter classes makes this more complicated than it needs to be.  The code might be simpler overall if the FilterManager just accepted three lists of regular expressions in its constructor: exclude files, include files, and exclude dirs.

b2/sync/folder.py
```diff
@@ -157,6 +164,10 @@ def _walk_relative_paths(self, local_dir, b2_dir, reporter):
                 )
             local_path = os.path.join(local_dir, name)
             b2_path = join_b2_path(b2_dir, name)
+
+            if filters_manager is not None and filters_manager.exclude(local_path):
```
bwbeach: I think you need separate tests for files and directories.  A directory that matches an excludeFile regex should not be excluded, because that will also exclude all of the files under it.

test/test_sync.py
```diff
@@ -105,6 +115,82 @@ def test_invalid_permissions(self):
...
+            six.u('inner/a.bin'),
+            six.u('inner/a.txt'),
+            six.u('inner/b.bin'),
+            six.u('inner/b.txt'),
+            six.u('\u81ea\u7531'),
+        ]
+        filters_manager = FilterManager(
+            [
+                ExcludeDirRegexFilter('.*hello'),
+                ExcludeDirRegexFilter('.*more'),
```
bwbeach: How about a test that an excludeFile regex doesn't exclude a directory?  Adding `ExcludeFileRegexFilter('innner')` would test that.

bwbeach: And also a test of the inverse would be good.  `ExcludeDirRegexFilter('hello0')` should not exclude the file `hello0`

test/test_sync.py
```diff
@@ -161,8 +247,11 @@ def __init__(self, f_type, files):
         self.f_type = f_type
         self.files = files
 
-    def all_files(self, reporter):
-        return iter(self.files)
+    def all_files(self, reporter, filters_manager=None):
+        for single_file in self.files:
+            if filters_manager is not None and filters_manager.exclude(single_file.name):
```
bwbeach: Maybe default `filters_manager` to a filter that includes all files?

b2/console_tool.py
```diff
@@ -767,6 +769,11 @@ class Sync(Command):
 
         Note that --includeRegex cannot be used without --excludeRegex.
 
+        You can specify --excludeDirRegex to selectively ignore whole directories
+        that match the given pattern. There is not other option, which overrides
+        --excludeDirRegex exclusions. The pattern is a regular expression
```
bwbeach: Instead of "There is not other option, which overrides --excludeDirRegex exclusions", how about:

```
When a directory is excluded, all of the files within it are excluded, even if they match an --includeRegex pattern.
```

b2/sync/scan_policies.py
```diff
@@ -0,0 +1,157 @@
...
+        self,
+        exclude_dir_regexes=tuple(),
+        exclude_file_regexes=tuple(),
+        include_file_regexes=tuple(),
+    ):
+        self._exclude_dir_polices = [ScanExcludeDirRegex(regex) for regex in exclude_dir_regexes]
+        self._include_file_polices = [ScanIncludeFileRegex(regex) for regex in include_file_regexes]
+        self._exclude_file_polices = [ScanExcludeFileRegex(regex) for regex in exclude_file_regexes]
+
+    def exclude(self, path):
```
bwbeach: I think it would be clearer to have separate methods for `should_exclude_file` and `should_exclude_directory`.  It is not obvious that this method handles both, and distinguishes based on whether the name has a trailing `/`.

b2/sync/folder.py
```diff
@@ -157,6 +163,10 @@ def _walk_relative_paths(self, local_dir, b2_dir, reporter):
                 )
             local_path = os.path.join(local_dir, name)
             b2_path = join_b2_path(b2_dir, name)
+
+            if policies_manager.exclude(local_path):
```
bwbeach: The check should be on the B2 name, not the local path.  Here's a test that fails because of this:

```
    def test_exclusion_with_exact_match(self):
        expected_list = [
            six.u('.dot_file'),
            six.u('hello.'),
            six.u('hello/a/1'),
            six.u('hello/a/2'),
            six.u('hello/b'),
            six.u('inner/a.bin'),
            six.u('inner/a.txt'),
            six.u('inner/b.bin'),
            six.u('inner/b.txt'),
            six.u('inner/more/a.bin'),
            six.u('inner/more/a.txt'),
            six.u('\u81ea\u7531'),
        ]
        polices_manager = ScanPoliciesManager(
            exclude_file_regexes=['hello0'],
        )
        self._check_file_filters_results(polices_manager, expected_list)
```

b2/sync/scan_policies.py
```diff
@@ -0,0 +1,157 @@
...
+
+class ScanExcludeDirRegex(ScanExcludePolicy):
+    def __init__(self, regex):
+        self.regex_str = regex
+        self.regex = re.compile(regex)
+
+    def __repr__(self):
+        return 'ExcludeDirRegex(%s)' % (self.regex_str,)
+
+    def should_exclude_file(self, file_path):
```
bwbeach: It seems like this method should not be needed.  If the entire directory was excluded when it was encountered, the code walking the directory should not be checking each file.

b2/sync/scan_policies.py
```diff
@@ -0,0 +1,157 @@
...
+        head, _ = os.path.split(file_path)
```
bwbeach: The names we're dealing with here are B2 names, which always use `/` as the separator.  `os.path.split` will want to use `\` if you're on Windows.

PR #411, opened 2018-02-15
Avoid making long list of futures when syncing lots of files.

bwbeach:


test/test_bounded_queue_executor.py
```diff
@@ -0,0 +1,71 @@
...
+
+    def test_two_exceptions(self):
+        def thrower():
+            raise Exception('test')
+
+        f1 = self.executor.submit(thrower)
+        f2 = self.executor.submit(thrower)
+        self.executor.shutdown()
+        self.assertEqual(2, self.executor.get_num_exceptions())
+        self.assertFalse(f1.exception() is None)
```
ppolewicz: The check here should be something more than `is not None` (for regression detection)

Issue #362, opened 2017-08-10
Upload error: File names must contain unicode characters with codes less than 32 - need indication of file


taz77:
I get a 400 error syncing files due to some file names that B2 does not like.
`UnknownError: Unknown error: 400 bad_request File names must not contain unicode characters with codes less than 32`

That is fine, but the problem is the b2 command line tool is not telling me which file it doesn't like so I have no way of fixing this (huge folders with over 300,000 files). The stdout should identify the filepath that caused the error so the user can correct the file name.

ppolewicz:
you are right, this should be fixed. For now, I think you could try enabling [detailed logs](https://github.com/Backblaze/B2_Command_Line_Tool#detailed-logs), it should give you some more information which will hopefully be sufficient to resolve the issue in your filesystem.

Znarkus:
Sadly --verbose doesn't help to identify which file is the problem

bwbeach:
One possibility is to validate the names before attempting to upload.  

The rules for file names are in the [B2 docs](https://www.backblaze.com/b2/docs/files.html).

dgentry:
This is fixed (by validating names before upload), although it could "try harder" in the face of an unusable filename.  The new `sync` behavior reports the offending filename in an exception (and stack trace), and stops execution.

The two "try harder" behaviors I can think of would be to:
1.  Report the offending filename and continue sync'ing everything else, or
2. Attempt to encode the offending filename into something usable, report the attempt, and continue.

But perhaps that should be opened in a separate issue.  The second one especially has additional details to work out.

bwbeach:
I agree that either of those two options should be a new issue, if somebody wants to argue for one of them.

I'm a little worried about changing the offending file name because it might change the sort order, so the the next sync would not match the source file with the one that was uploaded.

PR #424, opened 2018-03-16
Reject unusable filenames early (before the server does).

dgentry:
This solves Issue #362 by checking filenames and halting if it finds an unusable filename.  (We can open another issue for what the behavior should be in this case.)

It raises an UnusableFilename exception with appropriate detailed message for filenames that don't meet the rules.

If the problem with a filename is that it has a character code < 32, it includes the offending filename in the exception that it raises, with a "\x00" style hex code in place of the unprintable character(s).

There is a positive and negative unit test for each file rule.

### DIFF
--- a/b2/raw_api.py
+++ b/b2/raw_api.py
@@ -454,11 +454,13 @@ class B2RawApi(AbstractRawApi):
 
         :param filename: A proposed filename in utf-8.
         :return: None if the filename is usable."""
-        if len(filename) < 1:
+        decoded_filename = filename.decode("utf-8", 'ignore')
+        name_length = len(decoded_filename)
+        if name_length < 1:
             raise UnusableFileName("Filename must be at least 1 character.")
-        if len(filename) > 1024:
+        if name_length > 1024:
             raise UnusableFileName("Filename must be at most 1025 characters.")
-        lowest_unicode_value = ord(min(filename.decode('utf-8', 'ignore')))
+        lowest_unicode_value = ord(min(decoded_filename))
         if lowest_unicode_value < 32:
             message = "Filename \"{}\" contains code {}, which is less than 32.".format(
                 self.unprintable_to_question(filename), lowest_unicode_value
@@ -492,13 +494,8 @@ class B2RawApi(AbstractRawApi):
         :param data_stream: A file like object from which the contents of the file can be read.
         :return:
         """
-        # If the file_name is unusable, don't try to use it for the upload.
-        try:
-            self.check_b2_filename(file_name)
-        except UnusableFileName as e:
-            sys.stderr.write("\nUpload fails.  Exception is:\n{}\n".format(repr(e)))
-            sys.stderr.write("Exiting.\n")
-            os._exit(1)
+        # This will raise UnusableFileName if the file_name doesn't meet the rules.
+        self.check_b2_filename(file_name)
         headers = {
             'Authorization': upload_auth_token,
             'Content-Length': str(content_length),
--- a/b2/raw_api.py
+++ b/b2/raw_api.py
@@ -17,7 +17,7 @@ import random
 import re
 import sys
 import time
-from abc import (ABCMeta, abstractmethod)
+from abc import ABCMeta, abstractmethod
 
 import six
 
@@ -33,6 +33,14 @@ SRC_LAST_MODIFIED_MILLIS = 'src_last_modified_millis'
 
 # Special X-Bz-Content-Sha1 value to verify checksum at the end
 HEX_DIGITS_AT_END = 'hex_digits_at_end'
+
+if six.PY3:
+    unichr = chr
+
+# Unicode characters for testing filenames.  (0x0394 is a letter Delta.)
+TWO_BYTE_UNICHR = unichr(0x0394)
+CHAR_UNDER_32 = unichr(31)
+DEL_CHAR = unichr(127)
 
 
 @six.add_metaclass(ABCMeta)
@@ -442,9 +450,14 @@ class B2RawApi(AbstractRawApi):
             **kwargs
         )
 
-    def unprintable_to_question(self, string):
-        """Replace unprintable chars with a question mark."""
-        return re.sub(r'[^\x20-\x7E]', '?', string)
+    def unprintable_to_hex(self, str):
+        """Replace unprintable chars with a hex representation."""
+        unprintables_pattern = re.compile(r'[\x00-\x1f]')
+
+        def hexify(match):
+            return r'\x{0:02x}'.format(ord(match.group()))
+
+        return unprintables_pattern.sub(hexify, str)
 
     def check_b2_filename(self, filename):
         """
@@ -452,18 +465,18 @@ class B2RawApi(AbstractRawApi):
 
         See https://www.backblaze.com/b2/docs/files.html for the rules.
 
-        :param filename: A proposed filename in utf-8.
+        :param filename: A proposed filename in unicode.
         :return: None if the filename is usable."""
-        decoded_filename = filename.decode("utf-8", 'ignore')
-        name_length = len(decoded_filename)
-        if name_length < 1:
+        encoded_name = filename.encode('utf-8')
+        length_in_bytes = len(encoded_name)
+        if length_in_bytes < 1:
             raise UnusableFileName("Filename must be at least 1 character.")
-        if name_length > 1024:
-            raise UnusableFileName("Filename must be at most 1025 characters.")
-        lowest_unicode_value = ord(min(decoded_filename))
+        if length_in_bytes > 1024:
+            raise UnusableFileName("Filename can be at most 1024 bytes.")
+        lowest_unicode_value = ord(min(filename))
         if lowest_unicode_value < 32:
-            message = "Filename \"{}\" contains code {}, which is less than 32.".format(
-                self.unprintable_to_question(filename), lowest_unicode_value
+            message = "Filename \"{}\" contains code {} (hex {:02x}), less than 32.".format(
+                self.unprintable_to_hex(filename), lowest_unicode_value, lowest_unicode_value
             )
             raise UnusableFileName(message)
         # No DEL for you.
@@ -473,9 +486,9 @@ class B2RawApi(AbstractRawApi):
             raise UnusableFileName("Filename may not start or end with '/'.")
         if '//' in filename:
             raise UnusableFileName("Filename may not contain \"//\".")
-        longest_segment = max([len(segment) for segment in filename.split('/')])
-        if longest_segment > 250:
-            raise UnusableFileName("No filename segment may exceed 250 characters.")
+        long_segment = max([len(segment.encode('utf-8')) for segment in filename.split('/')])
+        if long_segment > 250:
+            raise UnusableFileName("No filename segment may exceed 250 bytes in utf-8.")
 
     def upload_file(
         self, upload_url, upload_auth_token, file_name, content_length, content_type, content_sha1,
@@ -525,7 +538,7 @@ def test_raw_api():
     """
     Exercises the code in B2RawApi by making each call once, just
     to make sure the parameters are passed in, and the result is
-    passed back.
+    passed back.  Also tests the filename rule checker
 
     The goal is to be a complete test of B2RawApi, so the tests for
     the rest of the code can use the simulator.
@@ -536,6 +549,7 @@ def test_raw_api():
     """
     try:
         raw_api = B2RawApi(B2Http())
+        test_raw_api_filename_rules(raw_api)
         test_raw_api_helper(raw_api)
         return 0
     except Exception as e:
@@ -609,13 +623,13 @@ def test_raw_api_helper(raw_api):
     print('b2_download_file_by_id (auth)')
     download_dest = DownloadDestBytes()
     raw_api.download_file_by_id(download_url, account_auth_token, file_id, download_dest)
-    assert file_contents == download_dest.bytes_io.getvalue()
+    assert file_contents == download_dest.get_bytes_written()
 
     # b2_download_file_by_id no auth
     print('b2_download_file_by_id (no auth)')
     download_dest = DownloadDestBytes()
     raw_api.download_file_by_id(download_url, None, file_id, download_dest)
-    assert file_contents == download_dest.bytes_io.getvalue()
+    assert file_contents == download_dest.get_bytes_written()
 
     # b2_download_file_by_name with auth
     print('b2_download_file_by_name (auth)')
@@ -623,13 +637,13 @@ def test_raw_api_helper(raw_api):
     raw_api.download_file_by_name(
         download_url, account_auth_token, bucket_name, file_name, download_dest
     )
-    assert file_contents == download_dest.bytes_io.getvalue()
+    assert file_contents == download_dest.get_bytes_written()
 
     # b2_download_file_by_name no auth
     print('b2_download_file_by_name (no auth)')
     download_dest = DownloadDestBytes()
     raw_api.download_file_by_name(download_url, None, bucket_name, file_name, download_dest)
-    assert file_contents == download_dest.bytes_io.getvalue()
+    assert file_contents == download_dest.get_bytes_written()
 
     # b2_get_download_authorization
     print('b2_get_download_authorization')
@@ -644,7 +658,7 @@ def test_raw_api_helper(raw_api):
     raw_api.download_file_by_name(
         download_url, download_auth_token, bucket_name, file_name, download_dest
     )
-    assert file_contents == download_dest.bytes_io.getvalue()
+    assert file_contents == download_dest.get_bytes_written()
 
     # b2_list_file_names
     print('b2_list_file_names')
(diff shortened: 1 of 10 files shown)


dgentry:
Here's a sample run where it rejects a file with a BEL/ctrl-G character in the name.

```
$ cp ~/be^Gll.txt test/
$ b2 sync test b2://saywhat
 compare: 1/18 files   updated: 0 files   0 B   0 B/s
Upload fails.  Exception is:
UnusableFileName('Filename "be?ll.txt" contains code 7, which is less than 32.',)
Exiting.
```

dgentry:
Fixed formatting so Travis is happier.

b2/raw_api.py
```diff
@@ -440,6 +441,38 @@ def update_bucket(
...
+        """
+        Raise an appropriate exception with details if the filename is unusable.
+
+        See https://www.backblaze.com/b2/docs/files.html for the rules.
+
+        :param filename: A proposed filename in utf-8.
+        :return: None if the filename is usable."""
+        if len(filename) < 1:
+            raise UnusableFileName("Filename must be at least 1 character.")
+        if len(filename) > 1024:
```
bwbeach: The limit is 1024 bytes (of UTF-8), not characters.

Here's a Unicode-in-Python learning opportunity.  :-)

b2/raw_api.py
```diff
@@ -440,6 +441,38 @@ def update_bucket(
...
+            raise UnusableFileName("Filename must be at most 1025 characters.")
+        lowest_unicode_value = ord(min(filename.decode('utf-8', 'ignore')))
+        if lowest_unicode_value < 32:
+            message = "Filename \"{}\" contains code {}, which is less than 32.".format(
+                self.unprintable_to_question(filename), lowest_unicode_value)
```
bwbeach: I'd be in favor of showing the hex value, maybe using `\xAB` notation.

dgentry: That was kind of what I was leaning towards, but wasn't sure if that was too fussy.  `ls` on my Mac just prints a `?` which isn't as helpful.  Anyway, I'll go ahead with the hex.

b2/raw_api.py
dgentry: Good catch!  I'll fix it.

b2/raw_api.py
```diff
@@ -457,6 +490,13 @@ def upload_file(
         :param data_stream: A file like object from which the contents of the file can be read.
         :return:
         """
+        # Raise an appropriate exception if the file_name is unusable.
+        try:
+            self.check_b2_filename(file_name)
+        except UnusableFileName as e:
+            sys.stderr.write("\nUpload fails.  Exception is:\n{}\n".format(repr(e)))
+            sys.stderr.write("Exiting.\n")
+            os._exit(1)
```
dgentry: This is where I think we should probably just report and skip the offending file instead of stopping dead.  Let me know if that would be better and I'll change it.  (I left it this way for ease of development, plus I don't yet know how to propagate a non-fatal error up to the overall sync status.)

bwbeach: It should just raise the exception.  The ConsoleTool class should catch it and report the error.

b2/raw_api.py
```diff
@@ -454,11 +454,13 @@ def check_b2_filename(self, filename):
 
         :param filename: A proposed filename in utf-8.
```
bwbeach: I should have noticed this before...

The name passed into `upload()` is unicode, not utf-8.  And that's what's passed in here.

b2/raw_api.py
```diff
@@ -454,11 +454,13 @@ def check_b2_filename(self, filename):
...
         :return: None if the filename is usable."""
-        if len(filename) < 1:
+        decoded_filename = filename.decode("utf-8", 'ignore')
```
bwbeach: Because the file name is already unicode, when I run this in Python 3, this line gets an error:

```
  File "/Users/brianb/sandbox/B2_Command_Line_Tool/b2/raw_api.py", line 457, in check_b2_filename
    decoded_filename = filename.decode("utf-8", 'ignore')
AttributeError: 'str' object has no attribute 'decode'
```

b2/raw_api.py
```diff
@@ -454,11 +454,13 @@ def check_b2_filename(self, filename):
...
+        name_length = len(decoded_filename)
+        if name_length < 1:
             raise UnusableFileName("Filename must be at least 1 character.")
-        if len(filename) > 1024:
+        if name_length > 1024:
```
bwbeach: The 1024 here is bytes, not characters.

b2/raw_api.py
```diff
@@ -454,11 +454,13 @@ def check_b2_filename(self, filename):
...
             raise UnusableFileName("Filename must be at most 1025 characters.")
```
bwbeach: Off by one.

ppolewicz:
Perhaps we should rather encode the filename and sync it successfully {maybe causing difficulty during recovery) rather than dropping it completely from sync and relying on the user to spot the error?

If I was using b2 cli for sync/backup, I'd really like ALL of my data to be synced.

If we encode it, we could add some Metadata to help us decode it back when syncing back.

dgentry:
Would you want metadata added per file?  Is there a natural place in b2 to
store that?

(I read about the 10 key-value metadata items b2 can store per file, which
might be a good place unless the the user is already using them. )

We could also do the old Windows-like 8.3 name mangling to just force the
file name to fit the b2 rules. They’re not super onerous. Besides files
with unprintable chars in the name, the case I would expect to hit
occasionally would be a sync of a super deep directory hierarchy.

I should be able to get back to this tomorrow at the latest.

bwbeach:
Encoding is an interesting idea.  Could the names be encoded in a way that preserves the sort order?

My original thought was just to raise an error and stop the sync, with a message that made it clear which file has an illegal name.

ppolewicz:
Let's wait for @bwbeach to comment before further coding

dgentry:
I went ahead and wrote test cases for each filename rule, since that should happen no matter what higher level sync behavior we turn out to choose.

The test cases are probably not where we ultimately want them (currently with `raw_test_api()`, which was broken for me, but which I think I fixed), but they and the filename checker worked at least once in python 2 and python 3.

Just for reference, here's the current higher level behavior when you try to sync a directory with one new bogus filename:

```
$ b2 sync test b2://saywhat
ERROR:b2.bucket:error when uploading, upload_url was https://pod-000-1103-14.backblaze.com/b2api/v1/b2_upload_file/bb20378e68e221dc60210817/c001_v0001103_t0058
  File "/Users/gentry/dg/B2_Command_Line_Tool/b2/bucket.py", line 413, in _upload_small_file
    HEX_DIGITS_AT_END, file_info, hashing_stream
  File "/Users/gentry/dg/B2_Command_Line_Tool/b2/raw_api.py", line 511, in upload_file
    self.check_b2_filename(file_name)
  File "/Users/gentry/dg/B2_Command_Line_Tool/b2/raw_api.py", line 481, in check_b2_filename
    raise UnusableFileName(message)
b2.exception.UnusableFileName: Unusable file name: Filename "be\x07ll.txt" contains code 7 (hex 07), less than 32.
ERROR:b2.sync.action:an exception occurred in a sync action
  File "/Users/gentry/dg/B2_Command_Line_Tool/b2/sync/action.py", line 42, in run
    self.do_action(bucket, reporter)
  File "/Users/gentry/dg/B2_Command_Line_Tool/b2/sync/action.py", line 84, in do_action
    progress_listener=SyncFileReporter(reporter)
  File "/Users/gentry/dg/B2_Command_Line_Tool/p3/lib/python3.6/site-packages/logfury-0.1.2-py3.6.egg/logfury/v0_1/trace_call.py", line 84, in wrapper
    return function(*wrapee_args, **wrapee_kwargs)
  File "/Users/gentry/dg/B2_Command_Line_Tool/b2/bucket.py", line 389, in upload
    return f.result()
  File "/usr/local/Cellar/python/3.6.4_4/Frameworks/Python.framework/Versions/3.6/lib/python3.6/concurrent/futures/_base.py", line 432, in result
    return self.__get_result()
  File "/usr/local/Cellar/python/3.6.4_4/Frameworks/Python.framework/Versions/3.6/lib/python3.6/concurrent/futures/_base.py", line 384, in __get_result
    raise self._exception
  File "/usr/local/Cellar/python/3.6.4_4/Frameworks/Python.framework/Versions/3.6/lib/python3.6/concurrent/futures/thread.py", line 56, in run
    result = self.fn(*self.args, **self.kwargs)
  File "/Users/gentry/dg/B2_Command_Line_Tool/b2/bucket.py", line 413, in _upload_small_file
    HEX_DIGITS_AT_END, file_info, hashing_stream
  File "/Users/gentry/dg/B2_Command_Line_Tool/b2/raw_api.py", line 511, in upload_file
    self.check_b2_filename(file_name)
  File "/Users/gentry/dg/B2_Command_Line_Tool/b2/raw_api.py", line 481, in check_b2_filename
    raise UnusableFileName(message)
b2.exception.UnusableFileName: Unusable file name: Filename "be\x07ll.txt" contains code 7 (hex 07), less than 32.
b2_upload(/Users/gentry/dg/B2_Command_Line_Tool/test/bell.txt, bell.txt, 1521188786128): UnusableFileName('Filename "be\\x07ll.txt" contains code 7 (hex 07), less than 32.',) Unusable file name: Filename "be\x07ll.txt" contains code 7 (hex 07), less than 32.
ERROR:b2.console_tool:ConsoleTool command error
  File "/Users/gentry/dg/B2_Command_Line_Tool/b2/console_tool.py", line 1098, in run_command
    return command.run(args)
  File "/Users/gentry/dg/B2_Command_Line_Tool/b2/console_tool.py", line 881, in run
    allow_empty_source=allow_empty_source
  File "/Users/gentry/dg/B2_Command_Line_Tool/p3/lib/python3.6/site-packages/logfury-0.1.2-py3.6.egg/logfury/v0_1/trace_call.py", line 84, in wrapper
    return function(*wrapee_args, **wrapee_kwargs)
  File "/Users/gentry/dg/B2_Command_Line_Tool/b2/sync/sync.py", line 220, in sync_folders
    raise CommandError('sync is incomplete')
b2.exception.CommandError: sync is incomplete
ERROR: sync is incomplete
```

b2/raw_api.py
dgentry: Fixed.

b2/raw_api.py
dgentry: Also fixed, since we are starting with unicode and utf-8 encoding it to get bytes.

b2/raw_api.py
dgentry: fixed.

b2/raw_api.py
dgentry: Good catch.  Fixed.

dgentry:
Oh, and to run the filename rule tests, run `b2 test-raw-api`.
```
$ b2 test-raw-api
b2 test filename rules
    Rule is:  Filename can be at most 1024 bytes.
    Rule is:  Filename can be at most 1024 bytes.
    Rule is:  Filename "hey \x1f joe" contains code 31 (hex 1f), less than 32.
    Rule is:  ASCII DEL (0x7f) not allowed.
    Rule is:  Filename may not start or end with '/'.
    Rule is:  Filename may not start or end with '/'.
    Rule is:  Filename may not contain "//".
    Rule is:  No filename segment may exceed 250 bytes in utf-8.
    Rule is:  No filename segment may exceed 250 bytes in utf-8.
b2_authorize_account
b2_create_bucket
.
.
.
b2_download_file_by_name (download auth)
Caught exception: InvalidAuthToken()
```
(which is further than it made it when it was failing on the missing `bytes_io` attribute.)

Like I said earlier, we probably ought to move these or set them up more like the other unit tests, but this at least shows them working, and already helped me find a bug in counting segment lengths with multi-byte characters.

bwbeach:
@dgentry - The code is looking good.  There's just one more change that would be nice to have: Use the unittest framework to exercise the file name checking, instead of including the tests in the `raw_api` module.

This project has two sets of tests: (1) unit tests, and (2) integration tests.  The `test_raw_api` is an integration test that exercises the connection to the actual B2 server.  The other integration test is `test_b2_command_line.py`.

The unit tests are in the test directory, in files named `test_*.py`.  Nose knows how to run them.  You can make a new `test_raw_api.py` file, and include the unit tests for `check_b2_filename`.

dgentry:
OK, moved the filename tests into the unit test framework.

b2/raw_api.py
```diff
@@ -570,27 +619,27 @@ def test_raw_api_helper(raw_api):
     print('b2_download_file_by_id (auth)')
     download_dest = DownloadDestBytes()
     raw_api.download_file_by_id(download_url, account_auth_token, file_id, download_dest)
-    assert file_contents == download_dest.bytes_io.getvalue()
+    assert file_contents == download_dest.get_bytes_written()
```
dgentry: I don't know if you'll want this change or not.  Without it, pre-commit.sh was failing here for me, but I don't recall failing integration tests in Travis before, so maybe something is different for my environment.  Either way, it kind of looks like the `DownloadDestBytes` author intended for `get_bytes_written()` to be used.

dgentry:
I hadn't been running `pre-commit.sh` because it seemed to be hanging (as yapf silently formatted all the files in the `.eggs` and any virtualenvs, as it turns out).

With these latest two commits, I am now running `pre-commit.sh` as my pre-commit hook.  They:
1.  Run yapf in parallel, and excluding the python under `.eggs/*`
2.  Modify the `make setup` target to enable `pre-commit.sh` as the pre-commit hook.

dgentry:
I'll add an integration test or two after lunch.  I figure I'll have b2 attempt a sync with just one kind of bad filename.  Is there anything else you think it should do?

test/test_raw_api.py
```diff
@@ -35,7 +36,7 @@ def _should_be_ok(self, filename):
         :param filename: unicode (or str) that follows the rules
         """
         print(u"Filename \"{0}\" should be OK".format(filename))
-        self.assertIsNone(self.raw_api.check_b2_filename(filename))
+        assert self.raw_api.check_b2_filename(filename) is None
```
bwbeach: It's better to use `self.assert*`.  These methods are part of the unittest framework, and report errors nicely.

Feel free to add `assertIsNone` to TestBase.  We have some other things there that weren't supported in 2.6.

pre-commit.sh
```diff
@@ -39,7 +39,7 @@ git pull --ff-only
 then checkout your topic branch and run $0.
 If the base branch on github is not called 'origin', invoke as $0 proper_origin_remote_name. Then your remote needs to be synched with your master too.
 """
-    yapf --in-place --recursive .
+    yapf --in-place --parallel --recursive --exclude '*eggs/*' .
```
bwbeach: Nice!

test/test_raw_api.py
```diff
@@ -0,0 +1,94 @@
...
+        From the B2 docs (https://www.backblaze.com/b2/docs/files.html):
+        - Names can be pretty much any UTF-8 string up to 1024 bytes long.
+        - No character codes below 32 are allowed.
+        - Backslashes are not allowed.
+        - DEL characters (127) are not allowed.
+        - File names cannot start with "/", end with "/", or contain "//".
+        - Maximum of 250 bytes of UTF-8 in each segment (part between slashes) of a file name.
+        """
+        print("test b2 filename rules")
+        self.raw_api = B2RawApi(B2Http())
```
bwbeach: This would normally go in the `setUp()` method in a unit test, which is used for constructor-ish things.  For each test method, the framework makes an instance of the test class, calls `setUp()`, and then calls the test method.  The other option would be to put this in the constructor.

It's nice to set all instance variables in either the constructor or setUp(); in a language without declarations, it helps document what the instance variables are.

The other option here would be to make `check_b2_filename` be a class method, and then not have to create an instance of `B2RawApi`.

test/test_raw_api.py
```diff
@@ -0,0 +1,94 @@
...
+        # 1024 bytes is ok if the segments are at most 250 chars.
+        s_1024 = 4 * (250 * 'x' + '/') + 20 * 'y'
+        self._should_be_ok(s_1024)
+        # 1025 is too long.
+        self._should_raise(s_1024 + u'x', "too long")
+        # 1024 bytes with two byte characters should also work.
+        s_1024_two_byte = 4 * (125 * TWO_BYTE_UNICHR + u'/') + 20 * u'y'
+        self._should_be_ok(s_1024_two_byte)
+        # But 1025 bytes is too long.
+        self._should_raise(s_1024_two_byte + u'x', "too long")
```
bwbeach: Glad you tested this case.  Chars vs. bytes always cause confusion, and this makes sure it's counting bytes.

bwbeach:
I think that additional integration test should do it.  Thanks for your work!

test/test_raw_api.py
```diff
@@ -0,0 +1,94 @@
...
+from __future__ import print_function
+
+import six
+
+from b2.raw_api import B2RawApi
+from b2.b2http import B2Http
+from b2.exception import UnusableFileName
+from .test_base import TestBase
+
+if six.PY3:
```
bwbeach: I don't think you need this hack.  `six.unichr()` is the function you want.

My PyCharm warns about redefining `unichr`.

dgentry:
My emacs elpy-mode complains about all kinds of other stuff, but not that.
I'll fix it.

test/test_base.py
```diff
@@ -14,6 +14,13 @@
 
 
 class TestBase(unittest.TestCase):
+
+    @contextmanager
```
bwbeach: `assertIsNone` is not a context manager, according to [the docs](https://docs.python.org/3/library/unittest.html#unittest.TestCase.assertIsNone).

test/test_base.py
```diff
@@ -14,10 +14,10 @@
 
 
 class TestBase(unittest.TestCase):
-
     @contextmanager
     def assertIsNone(self, expr):
         """Fail the test unless the expression is None."""
+        yield
```
bwbeach: Why?  

There's no mention of this is the checking message.

test/test_base.py
dgentry: Also takes another arg, `msg`.  Just checked the source to unittest2.

dgentry:
Removed the `yield`, since, as you pointed out, assertIsNone shouldn't be a
contextmanager/generator.  This change went with the previous commit
because I noticed the missing `yield` while I was waiting for the
pre-commit hook to finish.

dgentry:
The code quality tool identifies the problem (in test/test_base.py):

> Parameters differ from overridden 'assertIsNone' method
> `def assertIsNone(self, expr, msg=None):`

which a) is surprising because I thought those parameters should match, and b) causes me to reconsider overriding anything at all.

I think we should just use [`unittest2`](https://pypi.python.org/pypi/unittest2) (the backport to 2.4+ that includes all the unittest features of python 2.7) instead of re-implementing the three that we do (`assertIsNone()`, `assertRaises()`, and `assertRaisesRegexp()`).

One risk is that it could change your unit test error reporting on failures a little bit.

@bwbeach Any objection to eliminating the re-implementations of these three methods?  Then `TestBase` becomes trivial and I could also factor it out if you like.

PR #426, opened 2018-03-17
Added more unit tests for get-bucket --showSize.

toddATavail:
- Improved documentation to describe behavior with respect to hidden files, versions, and hierarchical structures.
- Renamed test_get_bucket_show_size() to test_get_bucket_one_item_show_size().
- Added test_get_bucket_with_versions() (to check behavior wrt to multiple versions).
- Added test_get_bucket_with_folders() (to check behavior wrt to folders).
- Added test_get_bucket_with_hidden() (to check behavior wrt to hidden files).
- Added test_get_bucket_complex() (to check everything together).

All tests pass.

### DIFF
--- a/b2/console_tool.py
+++ b/b2/console_tool.py
@@ -429,9 +429,12 @@ class GetBucket(Command):
 
         If --showSize is specified, then display the number of files
         (fileCount) in the bucket and the aggregate size of all files
-        (totalSize). Note that this entails additional API calls, so
-        will incur additional latency, computation, and Class C
-        transactions.
+        (totalSize). Hidden files are accounted for in the number of
+        files, but not in the aggregate size. Each version of a file
+        counts as an individual file, and its size contributes toward
+        the aggregate size. Analysis is recursive. Note that this
+        entails additional API calls, so will incur additional latency,
+        computation, and Class C transactions.
     """
 
     OPTION_FLAGS = ['showSize']
--- a/b2/console_tool.py
+++ b/b2/console_tool.py
@@ -429,11 +429,13 @@ class GetBucket(Command):
 
         If --showSize is specified, then display the number of files
         (fileCount) in the bucket and the aggregate size of all files
-        (totalSize). Hidden files are accounted for in the number of
-        files, but not in the aggregate size. Each version of a file
-        counts as an individual file, and its size contributes toward
-        the aggregate size. Analysis is recursive. Note that this
-        entails additional API calls, so will incur additional latency,
+        (totalSize). Hidden files and hide markers are accounted for
+        in the reported number of files, and hidden files also
+        contribute toward the reported aggregate size, whereas hide
+        markers do not. Each version of a file counts as an individual
+        file, and its size contributes toward the aggregate size.
+        Analysis is recursive. Note that --showSize requires multiple
+        API calls, and will therefore incur additional latency,
         computation, and Class C transactions.
     """
 
--- a/test/test_console_tool.py
+++ b/test/test_console_tool.py
@@ -527,7 +527,7 @@ class TestConsoleTool(TestBase):
         '''
         self._run_command(['get-bucket', '--showSize', 'my-bucket'], expected_stdout, '', 0)
 
-    def test_get_bucket_show_size(self):
+    def test_get_bucket_one_item_show_size(self):
         self._authorize_account()
         self._create_my_bucket()
         with TempDir() as temp_dir:
@@ -549,7 +549,7 @@ class TestConsoleTool(TestBase):
                 expected_stdout, '', 0
             )
 
-            # Now check the size information for the bucket.
+            # Now check the output of get-bucket against the canon.
             expected_stdout = '''
             {
                 "accountId": "my-account",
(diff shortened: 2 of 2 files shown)


test/test_console_tool.py
```diff
@@ -566,6 +566,183 @@ def test_get_bucket_show_size(self):
...
+        bucket.upload(UploadSourceBytes(b'test'), 'upload3')
+        bucket.upload(UploadSourceBytes(b'test'), 'upload4')
+        bucket.upload(UploadSourceBytes(b'test'), 'upload5')
+        bucket.upload(UploadSourceBytes(b'test'), 'upload6')
+
+        # Hide some new files. Don't check the results here; it will be clear enough that
+        # something has failed if the output of 'get-bucket' does not match the canon.
+        stdout, stderr = self._get_stdouterr()
+        console_tool = ConsoleTool(self.b2_api, stdout, stderr)
+        console_tool.run_command(['b2', 'hide_file', 'my-bucket', 'hidden1'])
```
bwbeach: When people hide files, the normal use case is to hide a file that already exists.  Uploading a hide marker for a file that does exist would be pointless.

In fact, it looks like this is a case where the simulator is inaccurate.  The real B2 service will refuse to hide a file that isn't there:

```
$ b2 hide-file bwb-test2 e
ERROR: File not present: e
```

test/test_console_tool.py
```diff
@@ -566,6 +566,183 @@ def test_get_bucket_show_size(self):
             '''
             self._run_command(['get-bucket', '--showSize', 'my-bucket'], expected_stdout, '', 0)
 
+    def test_get_bucket_with_versions(self):
+        self._authorize_account()
+        self._create_my_bucket()
+
+        # Put many versions of a file into the test bucket. Unroll the loop here for convenience.
```
bwbeach: Why unroll the loop?

test/test_console_tool.py
```diff
@@ -566,6 +566,183 @@ def test_get_bucket_show_size(self):
...
+        self._run_command(['get-bucket', '--showSize', 'my-bucket'], expected_stdout, '', 0)
+
+    def test_get_bucket_with_folders(self):
+        self._authorize_account()
+        self._create_my_bucket()
+
+        # Create a hierarchical structure within the test bucket. Unroll the loop here for
+        # convenience.
+        bucket = self.b2_api.get_bucket_by_name('my-bucket')
+        bucket.upload(UploadSourceBytes(b'test'), 'test')
```
bwbeach: Why so many files?  Not that it's a real problem, but I might have stopped at a depth of 3 or 4.  I think the point of this test is to make sure that it walks files inside folders, and once you're down a couple levels that probably proves the point.

test/test_console_tool.py
```diff
@@ -566,6 +566,183 @@ def test_get_bucket_show_size(self):
...
+            "corsRules": [],
+            "fileCount": 10,
+            "lifecycleRules": [],
+            "revision": 1,
+            "totalSize": 24
+        }
+        '''
+        self._run_command(['get-bucket', '--showSize', 'my-bucket'], expected_stdout, '', 0)
+
+    def test_get_bucket_complex(self):
```
bwbeach: Maybe add a comment saying what "complex" means?  Is this test meant to exercise versions, hiding, and folders at the same time?

b2/console_tool.py
```diff
@@ -429,9 +429,12 @@ class GetBucket(Command):
 
         If --showSize is specified, then display the number of files
         (fileCount) in the bucket and the aggregate size of all files
-        (totalSize). Note that this entails additional API calls, so
-        will incur additional latency, computation, and Class C
-        transactions.
+        (totalSize). Hidden files are accounted for in the number of
+        files, but not in the aggregate size. Each version of a file
```
bwbeach: I believe that hidden files are included in the total size.  It's the hide markers that have size 0 and are not included in the total size.

test/test_console_tool.py
toddATavail: The simulator thinks that hiding files creates new files. When I originally added `n` files and then hid `n` files, it thought that `2n` files were in existence. I don't have enough experience to identify differences between the simulator and the service a priori. I assumed that the simulator was "good enough". It sounds like this test exposes a problem with the simulator that should be addressed to improve verisimilitude.

bwbeach: Yep.  Not your fault the simulator is wrong.  I'm the one who wrote it.  :-/

Hiding a file adds a version of the file that is a "hide marker".  The original version is still there, and there is a zero-size hide marker.  The [docs](https://www.backblaze.com/b2/docs/file_versions.html) talk about this some.

test/test_console_tool.py
toddATavail: It's more convenient to structure the tests here through lexical code cloning rather than algorithmic duplication, especially in some of the tests that create varied hierarchical structures. While it's certainly possible to rewrite using loops and interpolation logic, I would argue that it's simpler and easier to maintain an unrolled loop. Given that it's Python and not a systems language, it's probably not any more efficient to unroll the loop; I presume that the VM overhead will dwarf any loop execution overhead, so efficiency isn't really a concern.

bwbeach: I agree that it's all a matter of code style: what's the easiest for a reader of the code to understand.  I'm fine with the code either way.

Efficiency is definitely not a concern here.  It's only doing a small number of files.

test/test_console_tool.py
toddATavail: It probably doesn't add anything to test 6 levels instead of 3, so the decision is somewhat arbitrary. Structural induction is probably a reasonable argument for stopping ~3-4, but hypothetically a strange implementation might introduce problems at some levels and not others. Problems are probably more likely to occur at enormous depths, if at all. Since all other things are essentially equal, 6 seemed as good as 3, so that's just what I happened to do.

test/test_console_tool.py
toddATavail: Maybe a better name, too: `test_get_bucket_show_size_all_features`?

b2/console_tool.py
toddATavail: My experimentation with service indicates that hidden files are not included in the total size, so the comment is accurate. The values produced by `--showSIze` also consistently agreed with the values reported by the website, in my experience. I can't comment on what it "should" do without a deeper understanding of the architects' intent, only what I observed it to actually do. I didn't test extensively, but I did smoke test ~6-7 data points, and they always agreed between the two mechanisms.

toddATavail: I should observe that I didn't try experiments with nearly as many files using my own real bucket and the real service. Just a couple files, which I would hide/unhide as necessary.

bwbeach: The totals on the web site com from the usage/billing system, which can lag by several minutes.

toddATavail: Okay, I'll retest and update the documentation accordingly.

bwbeach: I uploaded some test files.  You can see them all with the new `--recursive` flag on `ls`:

```
$ python -m b2 ls --versions --recursive --long bwb-test2
4_z4899f00dac54390c6e2f0519_f10552ec61b4a3769_d20180319_m193043_c000_v0001059_t0056    hide  2018-03-19  19:30:43          0  a
4_z4899f00dac54390c6e2f0519_f101aa4b02532b543_d20180316_m211136_c000_v0001058_t0053  upload  2018-03-16  21:11:36          1  a
4_z4899f00dac54390c6e2f0519_f112317c7cf6f77fa_d20180316_m211105_c000_v0001054_t0007  upload  2018-03-16  21:11:05          4  a
4_z4899f00dac54390c6e2f0519_f1162d927fa37bf30_d20180316_m211112_c000_v0001055_t0044  upload  2018-03-16  21:11:12          2  b
4_z4899f00dac54390c6e2f0519_f114f1f9f8e436775_d20180316_m211105_c000_v0001060_t0022  upload  2018-03-16  21:11:05          8  c/c1
4_z4899f00dac54390c6e2f0519_f105eb80bde615bc3_d20180316_m211112_c000_v0001060_t0018  upload  2018-03-16  21:11:12         16  c/c2
4_z4899f00dac54390c6e2f0519_f108cfa8045e6ebf4_d20180317_m004218_c000_v0001058_t0056  upload  2018-03-17  00:42:18          4  d
```

And the answer from your code is correct:

```
$ python -m b2 get-bucket --showSize bwb-test2
{
    "accountId": "890dc49cef59",
    "bucketId": "4899f00dac54390c6e2f0519",
    "bucketInfo": {},
    "bucketName": "bwb-test2",
    "bucketType": "allPrivate",
    "corsRules": [],
    "fileCount": 7,
    "lifecycleRules": [],
    "revision": 1,
    "totalSize": 35
}
```

toddATavail: Okay, so no changes are required to either my code or my documentation? If I'm understanding that correctly, do you still need me to change anything, or do you want to close your change request?

bwbeach:
I'd like the documentation updated to say that hidden files are included in the total size.  You could also add that hide markers count as files, but do not contribute to the size.

toddATavail:
Cool, thanks for clarifying. I'll take care of that in the next few minutes and resubmit.

toddATavail:
I pushed another change to clarify, per your request. Let me know if you like the new language. Here it is, for convenience:

```
    b2 get-bucket [--showSize] <bucketName>

        Prints all of the information about the bucket, including
        bucket info, CORS rules and lifecycle rules.

        If --showSize is specified, then display the number of files
        (fileCount) in the bucket and the aggregate size of all files
        (totalSize). Hidden files and hide markers are accounted for
        in the reported number of files, and hidden files also
        contribute toward the reported aggregate size, whereas hide
        markers do not. Each version of a file counts as an individual
        file, and its size contributes toward the aggregate size.
        Analysis is recursive. Note that --showSize requires multiple
        API calls, and will therefore incur additional latency,
        computation, and Class C transactions.
```

PR #436, opened 2018-04-10
Translate 404/not_found of get_file_info into FileNotPresent exception

elemental-lf:
Currently this return code gets translated into an UnknownError exception which is not very helpful

### DIFF
--- a/b2/exception.py
+++ b/b2/exception.py
@@ -348,10 +348,19 @@ def interpret_b2_error(status, code, message, post_params=None):
         return FileAlreadyHidden(post_params['fileName'])
     elif status == 400 and code == 'bad_json':
         return BadJson(message)
-    elif status == 400 and code in ("no_such_file", "file_not_present"):
-        # hide_file returns "no_such_file"
-        # delete_file_version returns "file_not_present"
-        return FileNotPresent(post_params['fileName'])
+    elif (
+        (status == 400 and code in ("no_such_file", "file_not_present")) or
+        (status == 404 and code == "not_found")
+    ):
+        # hide_file returns 400 and "no_such_file"
+        # delete_file_version returns 400 and "file_not_present"
+        # get_file_info returns 404 and "not_found"
+        # download_file_by_name/download_file_by_id return 404 and "not_found"
+        # but don't have post_params
+        if 'fileName' in post_params:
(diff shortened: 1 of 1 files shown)


b2/exception.py
```diff
@@ -348,9 +348,11 @@ def interpret_b2_error(status, code, message, post_params=None):
         return FileAlreadyHidden(post_params['fileName'])
     elif status == 400 and code == 'bad_json':
         return BadJson(message)
-    elif status == 400 and code in ("no_such_file", "file_not_present"):
-        # hide_file returns "no_such_file"
-        # delete_file_version returns "file_not_present"
+    elif ((status == 400 and code in ("no_such_file", "file_not_present"))
+        or (status == 404 and code in ("not_found"))):
+        # hide_file returns 400 and "no_such_file"
```
bwbeach: Thanks for the comments.  Very helpful.

elemental-lf:
@bwbeach Thank you for feedback. I'll correct this. But there are more problems unfortunately: Further testing showed that download_file_by_id and download_file_by_id can also return 404/not_found but they don't set post_params['fileName'] because the information is part of the request URL. I'll try to fix this also.

elemental-lf:
Regarding the AppVeyor test: The Python 3 job succeeds, but the Python 2 job fails with an error which I think isn't related to this PR:
```
Executing: python setup.py build
  File "setup.py", line 20, in <module>
    from setuptools import setup, find_packages, __version__ as setuptoolsversion
ImportError: No module named setuptools
Command exited with code 1``
```

bwbeach:
The AppVeyor build has been fragile.

I'll take a look and see what I can figure out.

bwbeach:
I just merged a PR to fix the appveyor build into `master`.  If you merge that into your branch, it should fix the build.

elemental-lf:
@bwbeach, hooray, this worked out okay. I'd really like this or a similar patch to be merged, because currently I have this in my code, which is ugly:
```python
isinstance(e, FileNotPresent) or isinstance(e, UnknownError) and "404 not_found" in str(e)
```
Thank you!

elemental-lf:
@bwbeach could this be merged please?

bwbeach:
Sorry for the delay merging.

PR #439, opened 2018-04-21
add contentSha1 to output of b2 upload-file

donovan:
This pull request includes the contentSha1 in upload-file the output. This is very useful for people who care about the integrity of the files they are uploading, current the workflow is this:

- compute local sha1sum
- call b2 upload-file
- parse output to get b2_file_id
- call  b2 get-file-info "$b2_file_id"
- parse get-file-info output to get remote sha1sum
- compare local and remote sha1sums

This patch simplifies this process such that the workflow is:

- compute local sha1sum
- call b2 upload-file
- parse upload-file output to get remote sha1sum
- compare local and remote sha1sums

### DIFF
--- a/b2/file_version.py
+++ b/b2/file_version.py
@@ -38,6 +38,8 @@ class FileVersionInfo(object):
             result['uploadTimestamp'] = self.upload_timestamp
         if self.action is not None:
             result['action'] = self.action
+        if self.content_sha1 is not None:
+            result['contentSha1'] = self.content_sha1
         return result
 
     def format_ls_entry(self):


svonohr:
Here is some background information on how uploading and file integrity checking works in B2 CLI. We have to differentiate between "large" (per default larger than 100MB) and "small" files.

Small files are uploaded in a single transaction. During the upload the SHA1 of the file is calculated and sent after the actual content of the file. The B2 server also calculates the SHA1 hash and verifies it against the hash sent by the client. The server’s response contains this hash which is verified again on the client side. This is the checksum your PR will display.

Large files are uploaded as smaller parts in multiple transactions. For each part the SHA1 is verified just as for small files. However, there is no SHA1 checksum calculated for the whole file by default. You can specify the whole SHA1 hash yourself using the `--sha1` option, but this is not calculated or verified by the server. So it can't be used to verify file integrity because it's always the hash you provided.

Regarding your PR, I think it could be useful to print the contentSha1 when it's available (which is only the case for small files). Anyway, verifying the file integrity using this hash is redundant, because this is already done by B2 CLI.

PR #451, opened 2018-06-29
Adding support for Application Key API's

spartan218:
Added the Create, Delete and List Key API's (not yet live) to the command availability.
Add test cases and some local checks in test_console_tool.py.

TODO: add functional testing in test_b2_command_line.py

Also fixed the issue we saw where the timestamp in milliseconds was returned differently for 
sync because we round in the unit tests, and not in sync (we no longer round), which we think
happened because the mac OSX changed the modified timestamp for files.

### DIFF
--- a/README.md
+++ b/README.md
@@ -25,7 +25,7 @@ this:
     b2 cancel-large-file <fileId>
     b2 clear-account
     b2 create-bucket [--bucketInfo <json>] [--corsRules <json>] [--lifecycleRules <json>] <bucketName> [allPublic | allPrivate]
-    b2 create-key [--duration <validDurationSeconds>] [--bucket <bucketId>] [--prefix <namePrefix>]
+    b2 create-key [--duration <validDurationSeconds>] [--bucket <bucketName>] [--prefix <namePrefix>] <capabilities> <keyName>
     b2 delete-bucket <bucketName>
     b2 delete-file-version [<fileName>] <fileId>
     b2 delete-key <applicationKeyId>
@@ -41,7 +41,7 @@ this:
     b2 list-buckets
     b2 list-file-names <bucketName> [<startFileName>] [<maxToShow>]
     b2 list-file-versions <bucketName> [<startFileName>] [<startFileId>] [<maxToShow>]
-    b2 list-keys [--keyCount <maxKeyCount>] [--startKey <startApplicationKeyId>]
+    b2 list-keys
     b2 list-parts <largeFileId>
     b2 list-unfinished-large-files <bucketName>
     b2 ls [--long] [--versions] [--recursive] <bucketName> [<folderName>]
--- a/README.md
+++ b/README.md
@@ -25,12 +25,14 @@ this:
     b2 cancel-large-file <fileId>
     b2 clear-account
     b2 create-bucket [--bucketInfo <json>] [--corsRules <json>] [--lifecycleRules <json>] <bucketName> [allPublic | allPrivate]
+    b2 create-key [--duration <validDurationSeconds>] [--bucket <bucketId>] [--prefix <namePrefix>]
     b2 delete-bucket <bucketName>
     b2 delete-file-version [<fileName>] <fileId>
+    b2 delete-key <applicationKeyId>
     b2 download-file-by-id [--noProgress] <fileId> <localFileName>
     b2 download-file-by-name [--noProgress] <bucketName> <fileName> <localFileName>
     b2 get-account-info
-    b2 get-bucket <bucketName>
+    b2 get-bucket [--showSize] <bucketName>
     b2 get-download-auth [--prefix <fileNamePrefix>] [--duration <durationInSeconds>] <bucketName>
     b2 get-download-url-with-auth [--duration <durationInSeconds>] <bucketName> <fileName>
     b2 get-file-info <fileId>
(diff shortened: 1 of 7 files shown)


spartan218:
I tried to enable Codacy for this build but it didn't let me. Not sure if it's a permissions issue or something else.

README.md
```diff
@@ -25,12 +25,14 @@ this:
     b2 cancel-large-file <fileId>
     b2 clear-account
     b2 create-bucket [--bucketInfo <json>] [--corsRules <json>] [--lifecycleRules <json>] <bucketName> [allPublic | allPrivate]
+    b2 create-key [--duration <validDurationSeconds>] [--bucket <bucketId>] [--prefix <namePrefix>]
```
bwbeach: All of the other commands that deal with buckets use bucketName.  This one should, too.

README.md
```diff
@@ -25,12 +25,14 @@ this:
...
     b2 delete-bucket <bucketName>
     b2 delete-file-version [<fileName>] <fileId>
+    b2 delete-key <applicationKeyId>
     b2 download-file-by-id [--noProgress] <fileId> <localFileName>
     b2 download-file-by-name [--noProgress] <bucketName> <fileName> <localFileName>
     b2 get-account-info
-    b2 get-bucket <bucketName>
+    b2 get-bucket [--showSize] <bucketName>
```
bwbeach: Was this an update we missed?

b2/console_tool.py
```diff
@@ -314,6 +314,41 @@ def run(self, args):
         return 0
 
 
+class CreateKey(Command):
+    """
+    b2 create-key [--duration <validDurationSeconds>] [--bucket <bucketId>] [--prefix <namePrefix>]
```
bwbeach: Later we can add a `--json` option to print the response JSON, instead of condensing to just the keyId and key.

README.md
```diff
@@ -39,14 +41,16 @@ this:
     b2 list-buckets
     b2 list-file-names <bucketName> [<startFileName>] [<maxToShow>]
     b2 list-file-versions <bucketName> [<startFileName>] [<startFileId>] [<maxToShow>]
+    b2 list-keys [--keyCount <maxKeyCount>] [--startKey <startApplicationKeyId>]
```
bwbeach: Maybe this should be like `b2 ls` and just list them all, without the `--keyCount` and `--startKey` options.

b2/console_tool.py
```diff
@@ -655,6 +705,33 @@ def run(self, args):
...
+    def run(self, args):
+        response = self.api.list_keys(
+            max_key_count=args.keyCount, start_application_key_id=args.startKey
+        )
+
+        self._print('Next Application Key Id: ' + response['nextApplicationKeyId'])
+
+        for key in response['keys']:
+            self._print(
+                '%s  %-10s  %s' % (key['applicationKeyId'], key['keyName'], key['capabilities'])
```
bwbeach: 10 is probably small for the key name

b2/raw_simulator.py
```diff
@@ -441,6 +441,8 @@ class RawSimulator(AbstractRawApi):
 
     MIN_PART_SIZE = 200
 
+    MAX_DURATION_IN_SECONDS = 86400000
```
bwbeach: add a comment saying what the number is.   1000 days?

b2/raw_simulator.py
```diff
@@ -506,6 +508,34 @@ def create_bucket(
...
+        self, api_url, account_auth_token, account_id, capabilities, key_name,
+        valid_duration_seconds, bucket_id, name_prefix
+    ):
+
+        if not re.match(r'^[A-Za-z0-9-]{1,100}$', key_name):
+            raise BadJson('illegal key name: ' + key_name)
+        if valid_duration_seconds is not None:
+            if not re.match(r'^[0-9]+$', valid_duration_seconds):
+                raise BadJson('illegal duration number: ' + valid_duration_seconds)
+            if int(valid_duration_seconds
```
bwbeach: Sometimes yapf does terrible formatting.  :-(

b2/sync/folder.py
```diff
@@ -193,7 +193,7 @@ def _walk_relative_paths(self, local_dir, b2_dir, reporter, policies_manager):
                 # Check that the file still exists and is accessible, since it can take a long time
                 # to iterate through large folders
                 if is_file_readable(local_path, reporter):
-                    file_mod_time = int(round(os.path.getmtime(local_path) * 1000))
```
bwbeach: I would probably have done this as a separate pull request.  It's an independent bug.

test_b2_command_line.py
```diff
@@ -428,6 +428,7 @@ def basic_test(b2_tool, bucket_name):
     b2_tool.should_fail(
         ['authorize_account', sys.argv[1], bad_application_key], r'nvalid authorization'
     )
+    print('######### ' + sys.argv[2])
```
bwbeach: take this out?

README.md
spartan218: When I ran the tool this came out

bwbeach: That's what I thought.  When those previous changes went in, the README didn't get updated.

Thanks for fixing it!

README.md
spartan218: It's in the raw_api.py so I thought we would want to allow them to do this. 
I will take it out and you can decide some time.

`def list_keys(
        self,
        api_url,
        account_auth_token,
        account_id,
        max_key_count=None,
        start_application_key_id=None
    ):`

b2/console_tool.py
spartan218: You've mentioned this before but I dont see --json being used anywhere in the project. 
I dont think it would be hard to add, we could just take it in as an arg and then whenever 
a class Command is called check for an option --json flag then just use json.dumps.

README.md
bwbeach: I think the list-files and list-versions calls are useful only to people learning to call the APIS, they're not very useful for people who just want to list their files.  The same will be true for listing keys.  When people list their keys, they just want to list them all.

spartan218:
Made the changes you requested, please take a look when you have time.

b2/console_tool.py
```diff
@@ -707,27 +706,19 @@ def run(self, args):
...
-       - the first key to return when listing keys
     """
 
-    OPTION_ARGS = ['keyCount', 'startKey']
-
     def run(self, args):
-        response = self.api.list_keys(
-            max_key_count=args.keyCount, start_application_key_id=args.startKey
-        )
+        response = self.api.list_keys()
```
bwbeach: This will need to list *all* the keys, not just the ones returned from the first call.  It will need to keep calling until there is no `nextApplicationKeyId` in the response.

ppolewicz:
I worked out the Codacy build status indicator with their support, should work now.

Something happened to travis though?

spartan218:
Pushed the code for listing all keys, that should end this pull request once reviewed.

Thanks

Issue #443, opened 2018-06-05
Add support for Application Keys


bwbeach:
Backblaze will be adding support for multiple application keys per account, and for application keys with restricted access.

This command-line tool will need commands to create, list, and delete keys.

It will also need changes to let it authorize with restricted keys.

bwbeach:
With application keys restricted to a single bucket, you are not allowed to call `b2_list_buckets` without specifying the name or ID of the one bucket the key is allowed to use.

When you call `b2_authorize_account`, it returns a new field called `allowed`, with the list of things the key is allowed to do, and the ID if the bucket it's restricted to (if there is a restriction).  When there is a bucket restriction, I'm thinking it would be a good idea to go ahead and get the details of that bucket and store them with the account info.  Then, when a command needs a bucket, it can check whether it's the bucket that is allowed, and print a helpful error message if not.  (For security reasons, the service just returns "not_authorized" if you try to access the wrong bucket.)

ppolewicz:
I would leave that logic up to the server. If one day B2 server allows to modify the application keys so that they can mutate their bucket permissions or allow to authorize for many buckets, then old versions of the CLI should not return false information.

We should instead change the error messages so that they say something like "the provided application key is not allowed to access this bucket or this bucket does not exist".

I would also add a print for CLI and a return value for API to `b2_authorize_account` in the next major version, to display the contents of the `allowed` field.

bwbeach:
@apexo - Backblaze updated the B2 service yesterday with some changes that should fix this problem.  Let me know if it keeps happening.

eviljim:
@bwbeach this is still happening for me. I can re-authorize with the same key and b2 command line will work for an additional 24 hours, but the auth still expires and forces me to log back in again.

This happened after your post. Not sure on the exact timing, but I authorized probably sometime Saturday and it failed early AM US Pacific time; I just re-authorized with the same key and it is working again (but probably temporarily).

bwbeach:
That's a puzzle.  I'm running version 1.3.4, and was able to use the command-line-tool just now using the auth it got last week.  

Next time it happens, could you run `b2 list-buckets --debugLogs` and let us know what's in `b2_cli.log`?

apexo:
```
2018-08-21 18:25:58,880 20282   140272242427648 b2.api  DEBUG   calling B2Api.list_buckets(self=<b2.api.B2Api object at 0x7f93a8083400>, bucket_name=None)
[...]

PR #459, opened 2018-07-16
Support for the new Application Keys feature in B2 (#443)

bwbeach:


b2/account_info/abstract.py
```diff
@@ -137,3 +182,20 @@ def take_large_file_upload_url(self, file_id):
     @abstractmethod
     def clear_large_file_upload_urls(self, file_id):
         pass
+
+    @staticmethod
+    def allowed_is_valid(allowed):
+        """
+        Returns true iff the 'allowed' dict passed in is legal.
```
ppolewicz: iff

b2/account_info/abstract.py
```diff
@@ -137,3 +182,20 @@ def take_large_file_upload_url(self, file_id):
     @abstractmethod
     def clear_large_file_upload_urls(self, file_id):
         pass
+
+    @staticmethod
```
ppolewicz: - Guido van Rossum, python BDFL, [wrote](https://mail.python.org/pipermail/python-ideas/2012-May/014969.html) this about `staticmethods`: 
> They're basically an accident -- back in the Python 2.2 days when I was inventing new-style classes and descriptors, I meant to implement class methods but at first I didn't understand them and accidentally implemented static methods first. Then it was too late to remove them and only provide class methods.

Generally, whenever there is a temptation to use `@staticmethod`, just use `@classmethod` instead

b2/account_info/allowed_policy.py
```diff
@@ -0,0 +1,61 @@
...
+    # Special case for account info stored before capabilities were implemented.
+    # If that happened, it must have been authorized with an account master key,
+    # and everything will be allowed.
+    allowed = account_info.get_allowed()
+    if allowed is None:
+        return
+
+    # Check that the requested capability is in the allowed list.
+    # TODO: remove the check for 'all' after the service bug is fixed
+    if capability not in allowed['capabilities']:
```
ppolewicz: this assumes `'capabilities' in allowed`, but `allowed_is_valid()` above does not check that this is a thing

b2/account_info/in_memory.py
```diff
@@ -29,32 +30,54 @@ def inner(*args, **kwargs):
...
         super(InMemoryAccountInfo, self).__init__(*args, **kwargs)
+        self._clear_in_memory_account_fields()
 
     def clear(self):
-        self.set_auth_data(None, None, None, None, None, None, None)
-        self._buckets = {}
+        self._clear_in_memory_account_fields()
         return super(InMemoryAccountInfo, self).clear()
 
+    def _clear_in_memory_account_fields(self):
```
ppolewicz: 👍

b2/account_info/in_memory.py
```diff
@@ -29,32 +30,54 @@ def inner(*args, **kwargs):
...
         self._realm = realm
+        self._allowed = allowed
 
     def refresh_entire_bucket_name_cache(self, name_id_iterable):
         self._buckets = dict(name_id_iterable)
 
     def get_bucket_id_or_none_from_bucket_name(self, bucket_name):
         return self._buckets.get(bucket_name)
 
+    def get_bucket_name_from_allowed_or_none(self):
```
ppolewicz: it doesn't look like this function can return `None`. At least comment it maybe?

b2/account_info/in_memory.py
```diff
@@ -89,3 +112,16 @@ def get_minimum_part_size(self):
     @_raise_missing_if_result_is_none
     def get_realm(self):
         return self._realm
+
+    def get_allowed(self):
+        return self._allowed
+
+    @_raise_missing_if_result_is_none
+    def get_allowed_bucket_id(self):
+        allowed = json.loads(self._allowed)
```
ppolewicz: this will raise a json deserialization error if constructor is called without specifying non-mandatory `allowed`

b2/account_info/sqlite_account_info.py
```diff
@@ -32,6 +32,9 @@ class SqliteAccountInfo(UrlPoolAccountInfo):
     """
     Stores account information in an sqlite database, which is
     used to manage concurrent access to the data.
+
+    The 'update_done' table tracks the schema updates that have been
```
ppolewicz: I like this very much, but it shouldn't be the part of the `keys` PR. It's an improvement for the next time.

b2/account_info/sqlite_account_info.py
```diff
@@ -161,6 +172,29 @@ def _create_tables(self, conn):
...
+    def _ensure_update(self, update_number, update_command):
+        """
+        Runs the update with the given number if it hasn't been done yet.
+
+        Does the update and stores the number as a single transaction,
+        so they will always be in sync.
+        """
+        with self._get_connection() as conn:
+            conn.execute('BEGIN')
+            cursor = conn.execute(
```
ppolewicz: `INSERT OR IGNORE INTO update_done (update_number) VALUES (?);` if the database has a feature for this, we should not be reimplementing it here

b2/account_info/sqlite_account_info.py
```diff
@@ -169,23 +203,31 @@ def clear(self):
...
-        application_key, realm
+        self,
+        account_id,
+        account_auth_token,
+        api_url,
+        download_url,
+        minimum_part_size,
+        application_key,
+        realm,
+        allowed=None
```
ppolewicz: missing trailing comma

b2/account_info/sqlite_account_info.py
```diff
@@ -169,23 +203,31 @@ def clear(self):
...
         with self._get_connection() as conn:
             conn.execute('DELETE FROM account;')
             conn.execute('DELETE FROM bucket;')
             conn.execute('DELETE FROM bucket_upload_url;')
             insert_statement = """
                 INSERT INTO account
-                (account_id, application_key, account_auth_token, api_url, download_url, minimum_part_size, realm)
-                values (?, ?, ?, ?, ?, ?, ?);
+                (account_id, application_key, account_auth_token, api_url, download_url, minimum_part_size, realm, allowed)
+                values (?, ?, ?, ?, ?, ?, ?, ?);
```
ppolewicz: `update_number INT NOT NULL` and this function has a default value of `None` for allowed argument, which will cause that insert to fail. Calling a function without specifying a non-mandatory argument, so that the default is used, should not cause an unconditional exception from the database layer...

b2/account_info/abstract.py
bwbeach: "iff" == "if and only if"

I'll spell it out.

b2/account_info/sqlite_account_info.py
bwbeach: I could not find that in sqlite.

b2/account_info/abstract.py
bwbeach: OK.  Good to know.

b2/account_info/sqlite_account_info.py
bwbeach: It's an improvement for this time, but could have been done before the app keys work.

b2/account_info/sqlite_account_info.py
bwbeach: I'm confused.  This insert statement is inserting a row with auth data. The update number is not part of this row.

b2/account_info/sqlite_account_info.py
bwbeach: Looks like I missed it.  But I don't think it's what we need here.  We need a transaction that atomically inserts the update number and does the update.  

I stand by the existing implementation.

b2/account_info/allowed_policy.py
bwbeach: Will add that check.

b2/account_info/in_memory.py
bwbeach: Better yet, remove the function.  :-)

b2/account_info/in_memory.py
bwbeach: ... and remove this function.

b2/account_info/sqlite_account_info.py
bwbeach: fixed

b2/api.py
```diff
@@ -123,12 +123,14 @@ def authorize_account(self, realm, account_id, application_key):
...
         response = self.raw_api.authorize_account(realm_url, account_id, application_key)
 
+        # TODO: make 'allowed' required once production servers start returning it
         self.account_info.set_auth_data(
             response['accountId'],
             response['authorizationToken'],
             response['apiUrl'],
             response['downloadUrl'],
             response['minimumPartSize'],
+            response.get('allowed'),
```
spartan218: It should be required, since I think production servers return this now...

b2/account_info/allowed_policy.py
```diff
@@ -0,0 +1,57 @@
...
+# Copyright 2018 Backblaze Inc. All Rights Reserved.
+#
+# License https://www.backblaze.com/using_b2_code.html
+#
+######################################################################
+
+from b2.exception import BucketNotAllowed, CapabilityNotAllowed, FileNameNotAllowed
+
+
+def check_command_allowed(capability, named_bucket, named_file, account_info):
```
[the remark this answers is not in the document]
ppolewicz: Actually, this is being used in a different manner and also has to be redesigned, performance should not be a concern in the new design

b2/account_info/sqlite_account_info.py
bwbeach: From my reading, the explicit `BEGIN` is necessary for some reason.  The transaction gets committed automatically at the end of the `with` block.

ppolewicz:
It's ok to do it after merging the PR, but before the release (console output might change in refactor and people can start relying on it)

b2/account_info/abstract.py
```diff
@@ -34,6 +34,17 @@ class AbstractAccountInfo(object):
         'staging': 'https://api.backblaze.net',
     }
 
+    # The 'allowed' structure to use for old account info that was saved without 'allowed'.
+    DEFAULT_ALLOWED = dict(
+        bucketId=None,
+        bucketName=None,
+        capabilities=[
```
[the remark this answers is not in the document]
ppolewicz: strange, I cannot find a good material on this. Maybe I'm wrong and it's not 2000. Lets keep it as lists for the sake of `json` serializer.

Issue #461, opened 2018-07-18
Need better error messages from B2Api and Bucket when unauthorized because of app keys


bwbeach:
When using an application key that's restricted to a bucket, the service returns a 403 Unauthorized that does not have a useful error message.  

There is code in `ConsoleTool` to attempt to provide useful information when unauthorized.  It would be better to move that down to the library layer so that library users also see the benefits.

bwbeach:
One simple approach would be to extend `B2Session` to catch unauthorized and add information about the application key restrictions.  The session doesn't know what capability is needed, or what the file name or bucketId is, so it can't say exactly what the problem is.  But it could generate an error message like this:

```
Unauthorized call using application key 001f3544fc1e12e000000000a, which is restricted to listBuckets,listFiles,readFiles,shareFiles,writeFiles,deleteFiles
```

Or:

```
Unauthorized call using application key 001f3544fc1e12e000000000a, which is restricted to 'listBuckets,listFiles,readFiles,shareFiles,writeFiles,deleteFiles' and is restricted to bucket 'my-bucket'
```

PR #462, opened 2018-07-19
Error reporting for application key authorization problems (#461)

bwbeach:


b2/raw_simulator.py
```diff
@@ -15,15 +15,32 @@
...
+    NonExistentBucket,
+    Unauthorized,
 )
 from .raw_api import AbstractRawApi, HEX_DIGITS_AT_END
 from .utils import b2_url_encode
 
 ALL_CAPABILITES = [
-    'listKeys', 'writeKeys', 'deleteKeys', 'listBuckets', 'writeBuckets', 'deleteBuckets',
-    'listFiles', 'readFiles', 'shareFiles', 'writeFiles', 'deleteFiles'
+    'listKeys',
```
ppolewicz: a clone, consider `import .account_info.abstract.ALL_CAPABILITIES` instead of redefining it

bwbeach: I debated that, but put off doing it because I wasn't sure that the `raw_api` should know about `account_info`.  

If anything, I think that this list belongs in the `raw_api`, and the account_info should use it from there.

ppolewicz: It should live in one place. I agree that `raw_api` is a better home for it.

PR #467, opened 2018-07-23
fix tests with Python 3.7

mwhudson:
PEP 479 changed behavior so that StopIteration does not bubble out of a generator.
Luckily the affected generators can just return immediately instead.


mwhudson:
This looks sane to me but I'm just looking at this code for the first time ever to get it working with Python 3.7 so please be skeptical :)

bwbeach:
Thanks!  I learned something new about Python today.

PR #468, opened 2018-07-23
Fix raw_api_test

bwbeach:
There were three problems:

1. The download authorization with file prefix did not match the file being downloaded.
2. The bucket revision test was too picky.  The only promise that the service makes is that updates to a bucket increase the revision number.
3. `pre-commit.sh` was not detecting errors from the `raw_api_test`

### DIFF
--- a/b2/console_tool.py
+++ b/b2/console_tool.py
@@ -1123,8 +1123,7 @@ class TestRawApi(Command):
     PRIVATE = True
 
     def run(self, args):
-        test_raw_api()
-        return 0
+        return test_raw_api()
 
 
 class TestUploadUrlConcurrency(Command):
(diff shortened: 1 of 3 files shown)


b2/console_tool.py
```diff
@@ -1123,8 +1123,7 @@ class TestRawApi(Command):
     PRIVATE = True
 
     def run(self, args):
-        test_raw_api()
-        return 0
```
ppolewicz: ouch

ppolewicz: should raise an exception though, it's the shell script that hid the error, right?

bwbeach: Command classes in `console_tool.py` return 0 or non-zero status.  And `test_raw_api` follows that convention.

I would be fine with changing `test_raw_api` to raise an exception if it sees a problem.

PR #470, opened 2018-07-26
List Bucket from allowed field only (if specified)

spartan218:
This is the fix for allowing us to make the CLI calls which require list buckets specifically when we have an application key that is restricted to a bucket.

### DIFF
--- a/b2/api.py
+++ b/b2/api.py
@@ -16,7 +16,7 @@ from .b2http import B2Http
 from .bucket import Bucket, BucketFactory
 from .cache import AuthInfoCache, DummyCache
 from .download_dest import DownloadDestProgressWrapper
-from .exception import NonExistentBucket
+from .exception import NonExistentBucket, RestrictedBucket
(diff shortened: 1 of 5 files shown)


ppolewicz:
Could you please add a test?

spartan218:
Yes, I was just telling Brian I should add a test.
Trying to decide whether or not to put the test in test_console_tool.py or make it 
a part of the integration tests. Thoughts @ppolewicz?

bwbeach:
We need both kinds of tests.  There should be a unit test that covers this, as well as a few integration tests that cover app keys and restricted app keys.  The integration tests don't need to cover all commands; the handling of app keys is not command specific.

ppolewicz:
I am talking to Codacy support about `Codacy was unable to analyse your pull request.`

also, Travis doesn't seem to have triggered here, I think due to missed hook, so it was a transient error and will resolve by itself when the branch is updated in any way. Hopefully...

spartan218:
Testing now included with some other small fixes we found along the way.

PR #471, opened 2018-07-27
Add information about required capabilities to doc string for each command.

bwbeach:
…mmand.


spartan218:
Part of me thinks that we should only let users authorize with keys that have the listBuckets capability for now that way users don't keep running into this issue.

bwbeach:
I like that idea.

PR #473, opened 2018-07-27
App key fixes

spartan218:
small fixes


spartan218:
@bwbeach Are these changes even needed still?

PR #480, opened 2018-08-01
Move download logic from raw_api to transferer

ppolewicz:
Fix some bugs on the way


b2/transferer.py
```diff
@@ -8,23 +8,104 @@
...
+                file_id,
+                file_name,
+                content_length,
+                content_type,
+                content_sha1,
+                file_info,
+                mod_time_millis,
+                range_=range_,
+            ) as file:
+                with progress_listener:
```
bwbeach: Seems like the DownloadDestProgressWrapper should take care of closing the progress listener.

ppolewicz: The progress bar situation is severe, I'll make a separate PR for fixing this

PR #483, opened 2018-08-02
Let devs pick a pre-commit integration test set

ppolewicz:


pre-commit.sh
```diff
@@ -127,18 +127,23 @@ function run_integration_tests
     fi
 }
 
+# get a list of tests to run
```
bwbeach: I'd put these up at the top, where the three existing arguments are "parsed".

Does this mean I cannot select tests without passing in the existing three (optional) arguments?

bwbeach: It's probably time for a usage comment at the top of the file.  Or a `--help` option.

ppolewicz: yes, in order to select tests you have to select other arguments as well. This was the case before too, if you wanted to use "quick" mode. I'll add a comment on the top.

PR #487, opened 2018-08-06
Faster downloads

ppolewicz:


b2/progress.py
```diff
@@ -218,10 +218,12 @@ def read(self, size=None):
         return data
 
 
-class StreamWithProgress(object):
+class AbstractStreamWithProgress(object):
     """
     Wraps a file-like object and updates a ProgressListener
-    as data is read and written.
+    as data is read / written.
+    In the abstract class, read and write methods do not update
```
bwbeach: Breaking out read and write streams is an improvement.  Now that it's clear that some streams are for reading and some for writing, it raises the question whether reading streams need a `write` method, and whether writing streams need a `read` method.   If not, it would clarify things even more.

b2/transferer.py
```diff
@@ -145,3 +230,79 @@ def as_info_dict(self):
...
+        self.url = response.request.url
+        self.writer = writer
+        self.range_ = range_
+        super(NonHashingDownloaderThread, self).__init__()
+
+    def run(self):
+        writer_queue_put = self.writer.queue.put
+        start_range = self.range_[0]
+        bytes_read = 0
+        with self.session.download_file_from_url(self.url, self.range_) as response:
```
bwbeach: When downloading a range of a large file, the offset in the local file is different than the offset in the B2 file.  I think this class needs to be clear about whether the range is relative to the file stored in B2 or the local file on disk.

b2/transferer.py
```diff
@@ -97,6 +118,70 @@ def _download_file_simple(self, file, response):
...
+        :param min_part_size: minimum amount of data a single stream will retrieve, in bytes
+        :return:
+        """
+        assert max_streams >= 1
+
+        info_dict = FileMetadata.from_response(response).as_info_dict()
+
+        content_length = info_dict['contentLength']
+        number_of_streams = min(max_streams, content_length // min_part_size)
+        part_size = content_length // number_of_streams
```
bwbeach: What happens if the content length is not a multiple of the number of streams?  The parts can't all be the same size.

b2/transferer.py
```diff
@@ -145,3 +230,79 @@ def as_info_dict(self):
...
+    def __enter__(self):
+        self.start()
+        return self
+
+    def __exit__(self, exc_type, exc_val, exc_tb):
+        self.queue.put((True, None, None))
+        self.join()
+
+
+class HashingDownloaderThread(threading.Thread):
```
bwbeach: This class differs in two ways from the NonHashing version: (1) it hashes, and (2) it stops reading before the end of the response body.  I think that's worth at least a comment, and maybe a different class name.  Maybe just call it `FirstPartDownloaderThread`

b2/transferer.py
bwbeach: I think the `_get_parts` method is asking for ranges that are one byte bigger that `part_size`, which would solve the problem.  It's probably worth some comments to make it clearer.

b2/progress.py
ppolewicz: They do:
in the new downloading implementation, the writing stream does some reading after everything is written, to complete hashing which has to be done start to end.

As they are now, progress listener wrappers can be chained to see both reading and writing from the stream separately. I would not like to take that away.

b2/transferer.py
ppolewicz: If I understand correctly, then this is an implementation detail of the b2 server and nobody outside of backblaze knows that the large files are stored in separate chunks. For the purpose of the client, remote offset and local offset, are identical.

This class assumes that byte N in the cloud will be byte N in the file. Is the assumption incorrect?

b2/transferer.py
ppolewicz: I'll fix it

ppolewicz:
Ok, please review, but do not merge yet, I'd like to improve the commit history when we finish iterating over this

b2/transferer.py
```diff
@@ -167,14 +167,21 @@ def _download_file_parallel(
...
+            (0, end_of_first_chunk),
+        )
         stream.start()
         streams = [stream]
 
-        start = part_size
+        start = end_of_first_chunk
         for part_number in range(1, number_of_streams):
-            end = (part_number + 1) * part_size
+            end = int((part_number + 1) * ideal_part_size)
```
bwbeach: Floating point numbers make me nervous.  Roundoff error could cause this truncation to be one less than expected.  On the other hand, I think there's an off-by-one error here because the end of a range is _inclusive_, which will try to make the end of the last part one higher than it should be.  When it downloads, though, HTTP says the server will clip the range to what's available in the file.

b2/transferer.py
bwbeach: I think that assumption is incorrect when downloading a range of a file.  

Suppose `Transferer.download_file_from_url` is asked to download the range (50, 69) of a 100-byte file, and decides to download the file in two parts.  The result will be a 20-byte file.  When it fetches the second part, it will need to ask for range (60, 69) from the service, and write those 10 bytes at offset 10 in the local file.

b2/transferer.py
ppolewicz: This needs to be fixed, the user might want to download a part of a file, then we'd retrieve one byte too much.

Do you have a suggestion on how to implement this properly?

bwbeach: I like math puzzles like this, so I was playing with it last night.  This is what I came up with:

```
    """
    HTTP ranges use an *inclusive* index at the end.
    """

    def __init__(self, start, end):
        assert 0 <= start <= end
        self.start = start
        self.end = end

    def size(self):
        return self.end - self.start + 1

    def subrange(self, sub_start, sub_end):
        """
        Returns a range that is part of this range.
        :param sub_start: Index relative to the start of this range.
        :param sub_end: (Inclusive!) index relative to the start of this range.
        :return: A new Range
        """
        assert 0 <= sub_start <= sub_end < self.size()
        return Range(self.start + sub_start, self.start + sub_end)

    def __repr__(self):
        return 'Range(%d, %d)' % (self.start, self.end)


class PartToDownload(object):
    """
    Holds the range of a file to download, and the range of the
    local file where it should be stored.
    """
    def __init__(self, cloud_range, local_range):
        self.cloud_range = cloud_range
        self.local_range = local_range

    def __repr__(self):
        return 'PartToDownload(%s, %s)' % (self.cloud_range, self.local_range)


def gen_parts(cloud_range, local_range, part_count):
    """
    Generates a sequence of PartToDownload to download a large file as
    a collection of parts.
    """
    assert cloud_range.size() == local_range.size()
    assert 0 < part_count <= cloud_range.size()
    offset = 0
    remaining_size = cloud_range.size()
    for i in range(part_count):
        # This rounds down, so if the parts aren't all the same size,
        # the smaller parts will come first.
        this_part_size = remaining_size // (part_count - i)
        yield PartToDownload(
            cloud_range.subrange(offset, offset + this_part_size - 1),
            local_range.subrange(offset, offset + this_part_size - 1)
        )
        offset += this_part_size
        remaining_size -= this_part_size
```

ppolewicz:
emailing Codacy support about the build status indicator (again...)

ppolewicz:
Another direction was chosen as preferred in #496 and others

PR #499, opened 2018-08-21
Fix reauthorizing with an application key

### LINKED ISSUE #498: Automatic re-authorizing with application key does not work
When it re-authorizes, it should use the same application key used for the original authentication, but it's using the account ID.

bwbeach:
Fix #498

### DIFF
--- a/b2/account_info/abstract.py
+++ b/b2/account_info/abstract.py
@@ -89,6 +89,10 @@ class AbstractAccountInfo(object):
         """ returns account_id or raises MissingAccountData exception """
 
     @abstractmethod
+    def get_account_id_or_app_key_id(self):
+        """ returns the account id or key id used to authenticate """
+
+    @abstractmethod
     def get_account_auth_token(self):
         """ returns account_auth_token or raises MissingAccountData exception """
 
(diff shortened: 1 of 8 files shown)


b2/account_info/abstract.py
```diff
@@ -176,8 +182,16 @@ def allowed_is_valid(cls, allowed):
...
+        self,
+        account_id,
+        auth_token,
+        api_url,
+        download_url,
+        minimum_part_size,
+        application_key,
+        realm,
+        allowed,
+        account_id_or_app_key_id,
```
spartan218: So one way to do this would have just been to save the appKeyId in the same area as accountId.
That way you dont have to add a new column and it should never be None. 

The other way is to make a new field.

Any thoughts on if just changing the name of account_id to account_id_or_app_key_id like the new field?  Is there a reason to keep just the accountId with an applicationKeyId?

b2/raw_simulator.py
```diff
@@ -930,6 +943,8 @@ def _assert_account_auth(
         assert key_sim is not None
         assert api_url == self.API_URL
         assert account_id == key_sim.account_id
+        if account_auth_token in self.expired_auth_tokens:
+            raise InvalidAuthToken('auth token expired', 'auth_token_expired')
```
spartan218: Same message twice?

PR #501, opened 2018-08-24
Fail download fast on invalid range

ppolewicz:


b2/exception.py
```diff
@@ -251,6 +251,19 @@ class UnusableFileName(B2SimpleError):
...
 
 
+class InvalidRange(B2Error):
+    def __init__(self, content_length, range_):
+        super(InvalidRange, self).__init__()
+        self.content_length = content_length
+        self.range_ = range_
+
+    def __str__(self):
+        return 'Cloud can only serve a range of 0-%d, while a range of %s was requested' % (
```
bwbeach: Thinking out loud because these kinds of things are tricky.  What about this case:

* File is 15 bytes long.
* Request is for range 10-19
* Response will have content length 5 and Content-Range of 10-14
* The message here will be wrong.

The Content-Range that comes back from the cloud will look like `10-14/15`.  That's probably the best place to get the actual file size.

test/test_bucket.py
```diff
@@ -462,55 +462,51 @@ def _make_data(self, approximate_length):
...
+        self.bucket.download_file_by_id(
+            self.file_info.id_, self.download_dest, self.progress_listener, range_=(0, 10)
+        )
+        self._verify('hello world')
 
     def test_download_by_id_progress_invalid_range(self):
-        file_info = self.bucket.upload_bytes(six.b('hello world'), 'file1')
-        download = DownloadDestBytes()
-        progress_listener = StubProgressListener()
         with self.assertRaises(InvalidRange):
```
bwbeach: Maybe check that the message is correct?

test/test_bucket.py
```diff
@@ -462,55 +462,51 @@ def _make_data(self, approximate_length):
...
 
 class DownloadTests(object):
-    def test_download_by_id_progress(self):
-        file_info = self.bucket.upload_bytes(six.b('hello world'), 'file1')
-        download = DownloadDestBytes()
-        progress_listener = StubProgressListener()
-        self.bucket.download_file_by_id(file_info.id_, download, progress_listener)
-        assert download.get_bytes_written() == six.b('hello world')
-        assert progress_listener.is_valid()
+    def setUp(self):
```
bwbeach: Nice!

b2/exception.py
ppolewicz: actually the message was correct. I extended tests to make it clear.

ppolewicz: This works, because content-length in such case will be actually 15

test/test_bucket.py
ppolewicz: implemented `msg` in our assertRaises and used it here

PR #502, opened 2018-08-26
Faster downloads prework7

ppolewicz:


test/test_bucket.py
```diff
@@ -512,17 +512,33 @@ def test_download_by_id_progress_range_one_off(self):
...
+            path = os.path.join(d, 'file2')
+            download_dest = PreSeekedDownloadDest(seek_target=3, local_file_path=path)
+            data = six.b('12345678901234567890')
+            write_file(path, data)
             self.bucket.download_file_by_id(
-                self.file_info.id_,
-                self.download_dest,
-                self.progress_listener,
-                range_=(5, 16),
+                self.file_info.id_, download_dest, self.progress_listener, range_=(3, 9)
```
bwbeach: trailing comma

b2/download_dest.py
```diff
@@ -59,6 +59,7 @@ class DownloadDestLocalFile(AbstractDownloadDestination):
     """
     Stores a downloaded file into a local file and sets its modification time.
     """
+    MODE = 'wb+'
```
bwbeach: How about a comment?  Something like "Truncate and open for writing"

b2/download_dest.py
```diff
@@ -97,6 +98,25 @@ def write_to_local_file_context(self, mod_time_millis):
             os.utime(self.local_file_path, (mod_time, mod_time))
 
 
+class PreSeekedDownloadDest(DownloadDestLocalFile):
+    """
+    Stores a downloaded file into a local file and sets its modification time.
+    Does not truncate the target file, seeks to a given offset just after opening
+    a descriptor.
+    """
+    MODE = 'rb+'
```
bwbeach: ... and a comment here: "open for updating, leaving existing contents"

test/test_bucket.py
```diff
@@ -512,17 +512,33 @@ def test_download_by_id_progress_range_one_off(self):
...
+        # hello world
+        #    |||||||
+        #    |||||||
+        #    vvvvvvv
+        #
+        # 123lo worl1234567890
+
+        with TempDir() as d:
+            path = os.path.join(d, 'file2')
+            download_dest = PreSeekedDownloadDest(seek_target=3, local_file_path=path)
```
bwbeach: It might be a better test if the local offset an the cloud offset were different.

b2/download_dest.py
ppolewicz: I dare not explain the full consequences of using that in a single-line comment. This, as a lot of other things we are using, is documented elsewhere.

Also, a comment here would make this line prone to being half-updated and thus not accurate, someone could easily lose data over this. I had consciously chosen to force the reader into looking it up, because it is too dangerous to attempt a shortcut.

test/test_bucket.py
ppolewicz: I had two tests there, might have not committed both. I'll check.

ppolewicz: Ok, I added another test with different offsets on local and cloud side

Issue #390, opened 2017-11-04
Needs option to not follow links


raindog308:
The 'b2 sync' command follows links and treats them as if they are simply directory paths.

It should either

- not follow links and report them as things it couldn't back up (which is what rclone does)

- follow them but give an option not to follow

Otherwise, a single ln command can radically alter backups.

I'm aware of exclude regex but that is a very crude tool and doesn't provide forward security...e.g., every time you symlink (or software you install unknowingly symlinks), you have to modify the backup script.

bwbeach:
This would be a good option to have.

Another thing that occurred to me is the possibility of storing symlinks in B2 by adding to the "file info" to say it's a symlink, and putting the contents of the link in the file.

Would you be interested, @raindog308, in looking at implementing this?  We have a group of friendly people on this project who would be happy to answer questions and offer advice.

dmanikowski-reef:
I can take it if @raindog308 does not want to :)

bhygate:
Dupe of #280?

Deffo needed though B-)

dmanikowski-reef:
We are almost done with it. Needed to do large code refactoring and there was a large discussion within our internal review.

claudio5678:
Nice to know it's almost done, can you provide an estimate date for it? thanks!

ppolewicz:
I think that got abandoned a few months ago, before it got finished...

joshlewis:
I just hit some circular symlinks deep inside a hard drive I'm trying to back up. That was a huge pain to discover and fix. Please give me the option to not follow them! 🙏

zackse:
@ppolewicz and @bwbeach can this issue be closed now that PR #512 is merged?

bwbeach:
Closing.  #512 fixed this problem.

PR #512, opened 2018-09-20
#390 Exclude symlinks from backup

zackse:
Opening this for comments -- I haven't visited this codebase in a while -- feedback on this approach is welcome and encouraged.

At first I thought of moving this into the `ScanPoliciesManager` routines, but since it affects both files and directories I elected to place the check in `LocalFolder._walk_relative_paths`.

```bash
zackse@goyeto# mkdir /tmp/symlink-test
zackse@goyeto# cd /tmp/symlink-test
zackse@goyeto# echo hello > hello
zackse@goyeto# ln -s hello hellolink
zackse@goyeto# mkdir adir
zackse@goyeto# echo hello > adir/hello2
zackse@goyeto# ln -s adir adirlink
zackse@goyeto# find -ls
126882253    0 drwxr-xr-x   6 zackse   wheel         204 Sep 20 13:18 .
126882256    0 drwxr-xr-x   3 zackse   wheel         102 Sep 20 13:18 ./adir
126882257    4 -rw-r--r--   1 zackse   wheel           6 Sep 20 13:18 ./adir/hello2
126882258    0 lrwxr-xr-x   1 zackse   wheel           4 Sep 20 13:18 ./adirlink -> adir
126882254    4 -rw-r--r--   1 zackse   wheel           6 Sep 20 13:18 ./hello
126882255    0 lrwxr-xr-x   1 zackse   wheel           5 Sep 20 13:18 ./hellolink -> hello

zackse@goyeto# b2 sync --delete --excludeSymlinks /tmp/symlink-test b2://mybucket/symlink-test
skipped symlink /tmp/symlink-test/adirlink
skipped symlink /tmp/symlink-test/hellolink
upload hello
upload adir/hello2

zackse@goyeto# b2 sync --delete /tmp/symlink-test b2://mybucket/symlink-test
upload hellolink
upload adirlink/hello2

zackse@goyeto# b2 sync --delete --excludeSymlinks /tmp/symlink-test b2://mybucket/symlink-test
skipped symlink /tmp/symlink-test/adirlink
skipped symlink /tmp/symlink-test/hellolink
delete adirlink/hello2
delete hellolink
```

### DIFF
--- a/b2/console_tool.py
+++ b/b2/console_tool.py
@@ -1016,7 +1016,7 @@ class Sync(Command):
             [--threads N] [--noProgress] [--dryRun ] [--allowEmptySource ] \\
             [--excludeRegex <regex> [--includeRegex <regex>]] \\
             [--excludeDirRegex <regex>] \\
-            [--excludeSymlinks ] \\
+            [--excludeAllSymlinks ] \\
             <source> <destination>
 
         Copies multiple files from source to destination.  Optionally
@@ -1057,7 +1057,7 @@ class Sync(Command):
 
         Note that --includeRegex cannot be used without --excludeRegex.
 
-        You can specify --excludeSymlinks to skip symlinks when
+        You can specify --excludeAllSymlinks to skip symlinks when
         syncing from a local source.
 
         When a directory is excluded by using --excludeDirRegex, all of
@@ -1137,7 +1137,7 @@ class Sync(Command):
 
     OPTION_FLAGS = [
         'delete', 'noProgress', 'skipNewer', 'replaceNewer', 'dryRun', 'allowEmptySource',
-        'excludeSymlinks'
+        'excludeAllSymlinks'
     ]
     OPTION_ARGS = ['keepDays', 'threads', 'compareVersions', 'compareThreshold']
     REQUIRED = ['source', 'destination']
@@ -1161,7 +1161,7 @@ class Sync(Command):
             exclude_dir_regexes=args.excludeDirRegex,
             exclude_file_regexes=args.excludeRegex,
             include_file_regexes=args.includeRegex,
-            exclude_symlinks=args.excludeSymlinks,
+            exclude_all_symlinks=args.excludeAllSymlinks,
         )
         sync_folders(
             source_folder=source,
--- a/b2/console_tool.py
+++ b/b2/console_tool.py
@@ -1136,8 +1136,13 @@ class Sync(Command):
     """
 
     OPTION_FLAGS = [
-        'delete', 'noProgress', 'skipNewer', 'replaceNewer', 'dryRun', 'allowEmptySource',
-        'excludeAllSymlinks'
+        'delete',
+        'noProgress',
+        'skipNewer',
+        'replaceNewer',
+        'dryRun',
+        'allowEmptySource',
+        'excludeAllSymlinks',
     ]
     OPTION_ARGS = ['keepDays', 'threads', 'compareVersions', 'compareThreshold']
     REQUIRED = ['source', 'destination']
--- a/b2/console_tool.py
+++ b/b2/console_tool.py
@@ -1136,7 +1136,8 @@ class Sync(Command):
     """
 
     OPTION_FLAGS = [
-        'delete', 'noProgress', 'skipNewer', 'replaceNewer', 'dryRun', 'allowEmptySource', 'excludeSymlinks'
+        'delete', 'noProgress', 'skipNewer', 'replaceNewer', 'dryRun', 'allowEmptySource',
+        'excludeSymlinks'
     ]
     OPTION_ARGS = ['keepDays', 'threads', 'compareVersions', 'compareThreshold']
     REQUIRED = ['source', 'destination']
--- a/b2/console_tool.py
+++ b/b2/console_tool.py
@@ -1016,6 +1016,7 @@ class Sync(Command):
             [--threads N] [--noProgress] [--dryRun ] [--allowEmptySource ] \\
             [--excludeRegex <regex> [--includeRegex <regex>]] \\
             [--excludeDirRegex <regex>] \\
+            [--excludeSymlinks ] \\
             <source> <destination>
 
         Copies multiple files from source to destination.  Optionally
@@ -1055,6 +1056,9 @@ class Sync(Command):
         of each file.
 
         Note that --includeRegex cannot be used without --excludeRegex.
+
+        You can specify --excludeSymlinks to skip symlinks when
+        syncing from a local source.
 
         When a directory is excluded by using --excludeDirRegex, all of
         the files within it are excluded, even if they match an --includeRegex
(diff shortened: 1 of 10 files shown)


ppolewicz:
Hi @zackse! I'm happy to discuss that as I've been scratching my head for a while with this issue.

I think we first need to understand why this feature is being implemented. If my understanding is correct, it is needed because of *circular* symlinks (ones that point to their parent directory) that impact the backup cycles negatively. Its not that all symlinks are bad though, some are actually good, backing them up may be important for some applications that rely on the symlink to be present and correct. Therefore if we merge this, one day a user may be faced with a choice:
1. Keep symlinks enabled and thus back up his source properly, so that the application that uses the underlying data can work immediately after restore, **but** a circular symlink will break the backups
2. Disable symlinks completely, **but** then application will not work after restore, because it relies on symlinks.

On a side note, I think the situation is even worse, because symlinks are not properly backed up / restored as it is now, so option 1 actually would create copy of a file the symlink points to.

I'm not saying the option to skip all symlinks is not needed. In fact, there are a few types of directory entries we probably want to skip by default, such as named fifos, devices etc and giving an ability to skip symlinks is a good start on that front... however it doesn't actually solve the actual problem of circular symlinks breaking backups. After this PR, backups are still going to be broken, just in a different way.

Perhaps instead (or in addition) of doing this we should detect circular symlinks and skip them?

zackse:
Hi @ppolewicz, thanks for your feedback.

Given the lack of support for symlinks (and presumably other special files, from your comments), the scope of work here could expand significantly... For a complete restore, we must have support for non-file/dir types, which means a plan for storing metadata vs. contents (e.g,. entry type, name, perms, target).

Since this doesn't exist yet, I agree that we should limit the scope of the work for issue #390 to the treatment of symlinks alone.

What do you think about the following two changes in the scope of #390, leaving the questions about (proper/full/restore-able) treatment of special types to another issue/ticket?

1. Skip circular symlinks (need more testing, but so far this appears to be limited to a symlink that points to the current or a parent directory), or skip any symlinks to directories
2. Add a new option `--excludeAllSymlinks` that implements the behavior in the draft of this pull request.

Thanks for working through this with me.

ppolewicz:
I think it is a solid plan, except that skipping circular symlinks should be implemented in a different PR than #512.

I don't think we can actually back up any other types than files, directories and symlinks. We should just not try to read named pipes, but lets leave that for later. Nobody is complaining about that, so lets not try to fix it.

Will you try to implement circular symlink detection?

As for #390, the very missing thing here is a test. Please add appropriate testing - we rely on that to ensure quality.

zackse:
Sounds good. I agree that separating the changes makes the most sense. I will rename the option `--excludeAllSymlinks` and add some tests.

As for circular detection, I can't commit to it just yet, but will try to review more of the code, as that would require some more record-keeping/state and thus be more intrusive than this changeset.

ppolewicz:
ok, lets track circular symlink breaking sync in #513

zackse:
@ppolewicz I'm not sure what to make of the build failure on Python 2.6, looks like an issue pulling in dependencies for running the test suite.

Let me know what you think of the changes to the option name and tests.

bwbeach:
I took a look at the build failure.  `twine` depends on `cmarkgfm`, which is no longer supported in 2.6.  I'll see about updating the requirements file to not depend on twine in 2.6, because twine is only needed when deploying to PyPI, and we don't use 2.6 for that.

bwbeach:
Just merged a build fix for 2.6 into master.  You can merge to pick it up.

test_b2_command_line.py
```diff
@@ -680,6 +680,37 @@ def _sync_test_using_dir(b2_tool, bucket_name, dir_):
...
+            expected_pattern="skipped symlink .*/alink"
+        )
+        file_versions = b2_tool.list_file_versions(bucket_name)
+        should_equal(
+            [
+                '+ ' + prefix + 'c',
+                '+ ' + prefix + 'c',
+                '+ ' + prefix + 'c',
+                '+ ' + prefix + 'linktarget',
+            ], file_version_summary(file_versions)
```
bwbeach: Optional style thing: Python allows a trailing comma in arg lists, and having one makes yapf format more nicely.  In this case, it would put `file_version_summary` on a new line.

bwbeach:
This all looks good to me.  Happy to merge, if you're ready.

b2/console_tool.py
```diff
@@ -1132,7 +1136,8 @@ class Sync(Command):
     """
 
     OPTION_FLAGS = [
-        'delete', 'noProgress', 'skipNewer', 'replaceNewer', 'dryRun', 'allowEmptySource'
+        'delete', 'noProgress', 'skipNewer', 'replaceNewer', 'dryRun', 'allowEmptySource',
+        'excludeAllSymlinks'
```
ppolewicz: please put a `,` after the last entry and re-run yapf, it will make a multi-line statement diff-friendly

b2/sync/report.py
```diff
@@ -198,6 +198,9 @@ def local_permission_error(self, path):
             'WARNING: %s could not be accessed (no permissions to read?)' % (path,)
         )
 
+    def local_symlink_skip_notice(self, path):
```
ppolewicz: event (non-progress) method names in this class are an answer to a question "what happened?", so "local access error", "local permission error", and here it should be "symlink skipped". Furthermore, in default implementation I think the method should be just `pass`: if the user specified an option to skip all symlinks, he doesn't expect to see names of all of those symlinks on his console.

test_b2_command_line.py
```diff
@@ -680,6 +680,37 @@ def _sync_test_using_dir(b2_tool, bucket_name, dir_):
...
+        b2_tool.should_succeed(['sync', '--noProgress', dir_path, b2_sync_point])
+        file_versions = b2_tool.list_file_versions(bucket_name)
+        should_equal(
+            [
+                '+ ' + prefix + 'alink',
+                '+ ' + prefix + 'c',
+                '+ ' + prefix + 'c',
+                '+ ' + prefix + 'c',
+                '+ ' + prefix + 'linktarget',
+            ], file_version_summary(file_versions)
```
ppolewicz: trailing comma would make multi-line formatting slightly nicer

test_b2_command_line.py
zackse: This differs from the style used in the prior calls to `should_equal`, is that OK?

test/test_console_tool.py
```diff
@@ -1059,6 +1059,41 @@ def test_sync_dry_run(self):
             ''' % (mtime)
             self._run_command(['list_file_names', 'my-bucket'], expected_stdout, '', 0)
 
+    def test_sync_exclude_all_symlinks(self):
+        self._authorize_account()
+        self._create_my_bucket()
+
+        with TempDir() as temp_dir:
+            file_path = os.path.join(temp_dir, 'test.txt')
+            with open(file_path, 'wb') as f:
```
ppolewicz: you should use `_make_local_file()` here

test/test_console_tool.py
```diff
@@ -1059,6 +1059,41 @@ def test_sync_dry_run(self):
...
+            ]
+            self._run_command(command, expected_stdout, '', 0)
+
+    def test_sync_dont_exclude_all_symlinks(self):
+        self._authorize_account()
+        self._create_my_bucket()
+
+        with TempDir() as temp_dir:
+            file_path = os.path.join(temp_dir, 'test.txt')
+            with open(file_path, 'wb') as f:
```
ppolewicz: you should use `_make_local_file()` here

zackse: Fixed.

test/test_console_tool.py
zackse: Fixed.

ppolewicz:
Looks very good now!

ppolewicz:
Thanks for your contribution!

PR #533, opened 2018-11-15
When a download fails, delete the partially-written local file.

bwbeach:


ppolewicz:
@bwbeach this is good, but please add a test case for PreSeekedDownloadDest. Without that test case, if one day PreSeekedDownloadDest stops inheriting from DownloadDestLocalFile, destination file might not get removed on failure and the tests will miss it. Lets skip the review of adding the test case and merge.

bwbeach:
I think that `TestPreSeekedDownloadDest` inherits the test case for deleting the file from `TestDownloadDestLocalFile`.

ppolewicz:
You are correct

Issue #547, opened 2019-02-02
This repository will be soon split in half


ppolewicz:
Hi everyone,

currently this repository holds the source code of two distinct projects:
- python client library for b2 cloud with supporting classes
- a CLI tool, which uses thaet client library

Those probably should have been split a long time ago, but that didn't happen. The current plan is to document the client library, put it [here](https://github.com/Backblaze/b2-sdk-python) and document it, so that python software developers can write tools which integrate with b2 with minimum effort.

The plan to do the move is as follows:
- [x] consult the plan with the community
- [x] push the whole code of the CLI to the SDK repository
- [x] remove the CLI from SDK code
- [x] release the SDK on pypi
- [x] remove SDK code from CLI, adding a dependency in the CLI from SDK instead, keeping the SDK classes which were accessible in namespace of the CLI, but providing a warning for anyone who tries to import those directly, so that users will update their dependencies
- [ ] close the SDK-related issues and PRs in the CLI. Their authors will be free to open a new issue in the SDK repository and reference the beginning of the discussion thread
- [ ] while we will be clearing the SDK issues, we will also close most of the stale CLI issues and PRs to clear the tracker from things which are not important, but make management of the project harder. Authors of the closed issues will be free to reopen them, if the issue is still relevant - it may sometimes be difficult for a reviewer to judge whether the problem has dissipated or not, so there will be some mistakes, but that's ok, since reopening can be done in a couple of clicks.
- [x] eventually with a top-version release of the CLI, we will remove the exposure of SDK code in the CLI package. Nobody is in a rush to do this, but eventually it is going to happen.

Please note that the plan does *not* have a stage in which the SDK is already separated, but there is no deprecation warning in the CLI when using the SDK code.

If anyone has a suggestion on how the plan outlined above may be improved, please speak up in the coming days.

whitebarry:
I like this idea a lot. I'm currently working on a thin client for an active project but it does seem very important to have an official standalone python library. Happy to help out any way I can.

ppolewicz:
Hi @whitebarry, we'll need someone to proofread and improve the public API documentation before release, could you help with that?

sww1235:
@ppolewicz I would also be willing to help out. I can certainly proofread as well.

whitebarry:
@ppolewicz Yeah absolutely happy to help on that.

jhart-r7:
Related questions:

[...]

PR #550, opened 2019-02-15
CLI functionality

dparamonov-reef:


ppolewicz:
This is a part of #547

bwbeach:
Some context in the PR description would be good.

It looks like this is step 5 of #547: "remove SDK code from CLI, adding a dependency in the CLI from SDK instead".  So I assume we won't be merging this to B2_Command_Line_Tool master branch until steps 1 through 4 have been completed.

bwbeach:
What's the strategy for testing that client code using the SDK from the existing `b2` package will continue to work, and be able to import the same things?

ppolewicz:
The plan outlined in #547 outdated quickly during implementation. The plan is to have one PR to the sdk repository and 3 PRs to CLI:

1. #550 with removing the SDK code
1. Next one to add warnings when someone attempts to import a sdk class from the CLI package
1. Another one to add tests which check if it's possible to import sdk classes from cli and to make sure that warnings are properly emitted if such import is attempted, but not when the ConsoleTool class is imported (as someone might want to subclass that and it shouldn't trigger a warning)

The scope of those 3 PRs was separated like that to make it easier to review.

bwbeach:
So this PR will not be merged until the SDK is in place and available on PyPI?

bwbeach:
Would it make sense to reorder the steps, so that client code using the SDK and importing it from the CLI package will continue to work with `master`?

1. Add tests for importing SDK classes from the CLI.
2. Switch CLI to use the SDK instead of the its own implementation.
3. Remove the SDK classes from the CLI.

ppolewicz:
> So this PR will not be merged until the SDK is in place and available on PyPI?
Correct, CI wouldn't pass otherwise.

> Would it make sense to reorder the steps (...)
If we have thought about those tests beforehand, then yes, that would be a very good plan. Unfortunately doing it now would require a lot of work due to a rebase impossible to resolve automatically. Therefore lets review the changes in the order they were done and just merge them all on the same day.

bwbeach:
Merging all at the same time will work.

PR #565, opened 2019-04-04
Correctly handle 0 byte file

oligriffiths:
This line causes an error for 0 byte files as `Range` asserts that both start and end are >= 0, thus if the file is 0 byes, -1 is passed in for end, causing an exception.


ppolewicz:
As the repository was split and affected code moved to another package, I'm moving your patch over to https://github.com/Backblaze/b2-sdk-python/pull/13

PR #586, opened 2019-06-29
Add support for b2 copy file

ppolewicz:


b2/console_tool.py
```diff
@@ -355,6 +355,100 @@ def run(self, args):
...
+
+        By default, it copies the file info and content type but you can replace
+        by setting the metadataDirective to 'replace'.
+
+        Content type is optional.  If not set, it will be set based on the
+        source file. It should only be provided when metadataDirective is replace and
+        should not be provided when metadataDirective is copy.
+
+        By default, the whole file gets copied but you can copy a range of bytes
+        from the source file to the new file.
```
bwbeach: Need to say whether the range is inclusive or exclusive.

b2/console_tool.py
```diff
@@ -355,6 +355,100 @@ def run(self, args):
         return 0
 
 
+class CopyFile(Command):
+    """
+    b2 copy-file [--metadataDirective [copy|replace]] [--contentType <contentType>] \\
+                 [--info <key>=<value>]* [--range start,end] \\
+                 <sourceFileId> <destinationBucketName> <b2FileName>
```
bwbeach: Do we want to a copy-by-name operation?

b2/console_tool.py
```diff
@@ -355,6 +355,100 @@ def run(self, args):
...
+        Content type is optional.  If not set, it will be set based on the
+        source file. It should only be provided when metadataDirective is replace and
+        should not be provided when metadataDirective is copy.
+
+        By default, the whole file gets copied but you can copy a range of bytes
+        from the source file to the new file.
+
+        info is optional. Each fileInfo is of the form "a=b".
+        If not set, it will be set based on the source file.
+        It should only be provided when metadataDirective is replace and
```
bwbeach: I would quote "replace" here.

b2/console_tool.py
```diff
@@ -355,6 +355,100 @@ def run(self, args):
         return 0
 
 
+class CopyFile(Command):
+    """
+    b2 copy-file [--metadataDirective [copy|replace]] [--contentType <contentType>] \\
```
bwbeach: This does not yet work for files over 5GB, right?  I don't see any code to handle large files, parts, etc.

ppolewicz: that's up to b2sdk, but yes, sdk 0.1.8 and sdk 0.1.10 can only handle small files. The PRs were 300 LOC each already, we wanted to merge that first before adding large file support.

bwbeach: I think it's worth documenting the limitation here.  Users of the CLI may not read the SDK docs.

b2/console_tool.py
ppolewicz: It would not be atomic. We can always easily add it later if requested, but I see value in the user retrieving the new file_id and then passing it here (in 2 lines of bash), as then the user and not b2cli, is responsible for dealing with concurrency. Also generally I think the amount of non-essential commands in cli should be kept to minimum, or we'll one day find ourselves in a situation where cli has too many commands for the user to comprehend.

b2/console_tool.py
ppolewicz: I think you might have reviewed an old version of this PR, I have already quoted it.

b2/console_tool.py
ppolewicz: I think you might have reviewed an old version of this PR, the current wording is `By default, the whole file gets copied but you can copy an (inclusive!) range of bytes from the source file to the new file.`

b2/console_tool.py
bwbeach: I guess I could make the same comment about the server-side API: specifying the source by name would be nice.  Since we don't do that, I think this interface is fine.

Do we want to leave room in the interface for names in the future.  We could have a `--sourceFileId` option, and then later add `--sourceFileName` and `--sourceBucket` options without breaking compatibility.  (Not sure that's a good idea.  Just thinking out loud.)

ppolewicz: if we do, then we should be consistent with downloads
```
b2 download-file-by-id [--noProgress] <fileId> <localFileName>
b2 download-file-by-name [--noProgress] <bucketName> <fileName> <localFileName>
```
so `copy-file-by-id` (and in the future maybe `copy-file-by-name` too (now that I think about it, that is a very small feature).

I'm convinced. Shall I change the name?

bwbeach: Yes.  Looks good to me.

PR #616, opened 2020-04-25
Move to b2sdk v1 interface

ppolewicz:


test/test_console_tool.py
```diff
@@ -227,16 +227,7 @@ def test_buckets(self):
...
-            "accountId": "{account_id}",
-            "bucketId": "bucket_1",
-            "bucketInfo": {{}},
-            "bucketName": "your-bucket",
-            "bucketType": "allPrivate",
-            "corsRules": [],
-            "lifecycleRules": [],
-            "revision": 1
-        }}
+        null
```
ppolewicz, line 230: @bwbeach do we bump b2cli version from 1.4.x to 1.5.0 due to this change potentially breaking interface? Nobody should be using this anyway, the return contents is not really documented, that's why it was removed... But I feel we should increase middle version. How do you think?

ppolewicz:
Travis almost passed here https://travis-ci.org/github/Backblaze/B2_Command_Line_Tool/builds/679431891 but yapf changes were not good (too new version of yapf on my workstation, I guess). After I removed that commit, travis didn't rebuild it, but I think we can say that it passed.

ppolewicz:
This works well when tested locally, but requires b2sdk 1.1.0 (not released yet, currently on `add_caches` branch) to work.

ppolewicz:
Unfortunately I need to merge it despite CI failures that it will trigger, since we will be mering a few more PRs, some of which may even need small changes in b2sdk. b2cli on master will have to rely on pre-commit script until we release b2sdk

PR #625, opened 2020-05-27
Use sphinx argparse

franciscouzo:
@ppolewicz


b2/console_tool.py
```diff
@@ -260,38 +207,45 @@ def __str__(self):
...
-        Requires capability: listBuckets
-    """
+    FORBID_LOGGING_ARGUMENTS = True
 
-    OPTION_FLAGS = ['dev', 'staging']  # undocumented
+    @classmethod
+    def add_subparser(cls, subparsers, parent):
+        parser = cls._add_parser(
+            subparsers,
+            """
```
ppolewicz, line 217: I think this should stay in `cls.__doc__` and it should just be referenced here. Same goes for every other Command. Is there any reason we need to remove the docstring from the class and keep it here? Also, in my design the code of `add_subparser` would be exactly the same for all classes, so it can use `super()` to run that line and then only add arguments

PR #632, opened 2020-06-03
Return json results more consistently, remove legacy commands

ppolewicz:


b2/console_tool.py
```diff
@@ -842,72 +840,24 @@ class ListBuckets(Command):
...
-
-@B2.register_subcommand
-class ListFileVersions(Command):
-    """
-    Lists the names of the files in a bucket, starting at the
-    given point.  This is a low-level operation that reports the
-    raw JSON returned from the service.  'b2 ls' provides a higher-
-    level view.  Optionally restricts the output (server-side) to the given prefix.
+    Alternatively, the --json option produces machine-readable output
+    similar to the server api response format.
```
ppolewicz, line 844:
```suggestion
    similar (but not identical) to the server api response format.
```

PR #633, opened 2020-06-04
Bundled version

mlech-reef:


.travis.yml
```diff
@@ -232,14 +282,39 @@ script:
...
+            if [[ "$TRAVIS_OS_NAME" == "windows" ]]; then B2_BINARY=b2.exe; else B2_BINARY=b2; fi
+            # create a bundle
+            pyinstaller --onefile b2.spec
+            # run integration tests
+            python test_b2_command_line.py --command=./dist/$B2_BINARY "$TEST_APPLICATION_KEY_ID" "$TEST_APPLICATION_KEY"
+            # upload the bundle to the B2 cloud
+            B2_VERSION=$(dist/$B2_BINARY version)
+            B2_TAG=$(git describe --tags --first-parent --always)
+            dist/$B2_BINARY authorize-account "$TEST_APPLICATION_KEY_ID" "$TEST_APPLICATION_KEY"
+            dist/$B2_BINARY upload-file --noProgress --info="b2_version=$B2_VERSION" --info="b2_tag=$B2_TAG" dep-b2-cli-"$TEST_APPLICATION_KEY_ID" dist/$B2_BINARY $TRAVIS_OS_NAME/$B2_BINARY
```
ppolewicz, line 313: I think we need to generate a download auth key that will let us, well, download the file, and we should echo the command that would need to be used to download the binary file in the end

mlech-reef: > I think we need to generate a download auth key that will let us, well, download the file
It's open-source, so the bucket used as storage can be `allPublic` in my opinion.

> and we should echo the command that would need to be used to download the binary file in the end
In my implementation, I have used account id inside the bucket name, so I can't print friendly URLs here as it will be masked by Travis. I can change the bucket name. Maybe we should ask Backblaze how they want to have it stored. Now, it will create a new file version each time and add B2 version and Git commit-ish id.

ppolewicz: I suggest to put it in a static bucket name (which does not contain account name) and then have `release/windows/b2.exe`, `release/linux/{64bit or glibc-version or something}/b2` etc. Echo the download link to stderr, wrapped by ############### header / footer, so that it's easy to find and extract from the bottom of the build log.

mlech-reef: But the bucket name must be globally unique, so if I create one for testing, I will have to delete it before we can create an official one, so in the future, it might be harder to test any changes for people who don't have write access to the Backblaze repo.
If it's ok, I can do it as you want. We may also pass it via env variable statically set up in Travis.

mlech-reef: Merging this PR must be coordinated with someone with access to Backblaze Travis.

mlech-reef: I'm gonna use `$TEST_BUCKET_NAME` env for that

ppolewicz: `TEST_BUCKET_NAME` does not properly rely the purpose of the bucket, we won't be using it for testing. How about `DEPLOY_BUCKET_NAME`?

ppolewicz:
We are just waiting for test results now

.travis.yml
mlech-reef: `DEPLOY_BUCKET_NAME` it is

ppolewicz:
We are merging this to save the code that generates the binaries etc, but travis is out of control, we will have to move away from it soon.

PR #635, opened 2020-06-17
Add aliases for commands_with_underscores

mlech-reef:


ppolewicz:
Where is travis when we need it...

PR #636, opened 2020-06-17
Fix modification time in one test

mlech-reef:


ppolewicz:
Regrettably travis runs tests pre-merge and not after-merge, that's how we landed with the failing tests on `master`.

PR #644, opened 2020-07-21
Drop Python 2 support

mlech-reef:


ppolewicz:
@bwbeach @mlech-reef should we drop `six` completely or should we keep it? In theory python 3.x line can have differences that `six` would help us cover gracefully, but that doesn't seem to happen from what I can see in six changelog for the past year and a half, nothing has really moved. It does affect performance though. I think it may be time to drop it.

ppolewicz:
https://github.com/Backblaze/b2-sdk-python/pull/146 is related

mlech-reef:
> @bwbeach @mlech-reef should we drop `six` completely or should we keep it? In theory python 3.x line can have differences that `six` would help us cover gracefully, but that doesn't seem to happen from what I can see in six changelog for the past year and a half, nothing has really moved. It does affect performance though. I think it may be time to drop it.

I think we should and I did some cleanup already in this PR and I probably can remove six completely from the CLI as the code is relatively small. But when it comes to SDK the code base is not so small, so maybe we should do it separately as there are probably lines of codes written especially for python2 and python3.4 which may don't even use six.

This is a blocker for new CI implementation as I want to use nox and pytest which don't get along with python 2.7.

ppolewicz:
we will either clean it up now when dropping 2.7/3.4 or it will never happen. I think it would be irresponsible to leave cruft, so unless there is a good reason to keep it (and it seems you and I are not aware of any) then we should get rid of it. Lets wait for Brian though, maybe there is something related to encoding (`six.b`) or something?

bwbeach:
I'm in favor of cleaning up the code and getting rid of `six`

mlech-reef:
Done

README.md
```diff
@@ -110,6 +104,8 @@ For advanced users, a hidden option `--logConfig <filename.ini>` can be used to
 
 ## Not released yet
 
+* Drop Python 2 support :tada:
```
ppolewicz, line 107:
```suggestion
* Drop Python 2 support :tada: (for old systems you can now use the [binary distribution](https://www.backblaze.com/b2/docs/quick_command_line.html))
```

PR #652, opened 2020-08-08
Clean up legacy code

mlech-reef:


ppolewicz:
I don't mind it staying (we'll probably need it if we b2sdk moves to v2), but we can put it in again when we need it.

PR #662, opened 2020-10-06
Remove nose and upgrade pytest

mlech-reef:


ppolewicz (approved):
nose won't work on 3.10, we have to get rid of it

PR #684, opened 2021-03-01
Facilitate for arrow's bug in parsing millis

mpnowacki-reef:
There's probably a bug in arrow: https://github.com/arrow-py/arrow/issues/933


mpnowacki-reef:
mistargeted PR

PR #694, opened 2021-03-17
Pin setuptools-scm<6.0

mlech-reef:


mlech-reef:
Approved internally by @mpnowacki-reef

PR #725, opened 2021-06-04
Add possibility to install SDK from local folder instead of pypi when running tests

mpnowacki-reef:


ppolewicz:
This was rebased and merged already

PR #746, opened 2021-09-11
Sign Windows bundle

mlech-reef:
Tested on fork:
image

### DIFF
--- a/noxfile.py
+++ b/noxfile.py
@@ -33,6 +33,9 @@ REQUIREMENTS_BUILD = ['setuptools>=20.2']
 OSX_BUNDLE_IDENTIFIER = 'com.backblaze.b2'
 OSX_BUNDLE_ENTITLEMENTS = 'contrib/macos/entitlements.plist'
 
+WINDOWS_TIMESTAMP_SERVER = 'http://timestamp.digicert.com'
+WINDOWS_SIGNTOOL_PATH = 'C:/Program Files (x86)/Windows Kits/10/bin/10.0.17763.0/x86/signtool.exe'
+
 nox.options.reuse_existing_virtualenvs = True
 nox.options.sessions = [
     'lint',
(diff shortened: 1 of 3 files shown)


mszumocki-reef (approved):

noxfile.py
```diff
@@ -200,30 +203,64 @@ def bundle(session):
...
-        '--entitlements',
-        OSX_BUNDLE_ENTITLEMENTS,
-        '--options',
-        'runtime',
-        *session.posargs,
-        'dist/b2',
-        external=True
-    )
-    session.run('codesign', '--verify', '--verbose', 'dist/b2', external=True)
+    if system == 'darwin':
```
ppolewicz, line 209: please refactor into sign_darwin, sign_windows

noxfile.py
```diff
@@ -33,6 +33,9 @@
 OSX_BUNDLE_IDENTIFIER = 'com.backblaze.b2'
 OSX_BUNDLE_ENTITLEMENTS = 'contrib/macos/entitlements.plist'
 
+WINDOWS_TIMESTAMP_SERVER = 'http://timestamp.digicert.com'
```
ppolewicz, line 36: this link doesn't work AND it's http?

noxfile.py
mlech-reef: How would it be better? It would be less consistent, for sure, as we don't have `bundle_darwin`, etc. and the binary can not be cross-signed anyway as in such case having per-OS command could make sense.

noxfile.py
mlech-reef: The command was provided by @ooola. I guess it uses https://www.ietf.org/rfc/rfc3161.txt and TSP over HTTP (chapter 3.4)

noxfile.py
ppolewicz: I mean instead of an enormous function with everything in it split signing for different operating systems into separate functions and then call them here

mlech-reef: Right ;)

noxfile.py
mlech-reef: HTTPS doesn't work with TSP - tested
image

ppolewicz (approved):
lgtm

ooola (approved):
ship it!

PR #747, opened 2021-09-11
Static binary for Linux

### LINKED ISSUE #724: Do not use Ubuntu 16.04 runner in CD as it is deprecated
Ubuntu 16.04 is now deprecated in GitHub Actions.
We use it now to create a binary with `pyinstaller` in order to use older `glibc` version so our bundle works even on older OSes (see [this](https://pyinstaller.readthedocs.io/en/stable/usage.html#making-gnu-linux-apps-forward-compatible) for details).

We can either:
* use Ubuntu 18.04 and drop support for OSes with older `glibc`
* use Docker runner with Ubuntu 16.04
* use `staticx` and create a fully-static binary

mlech-reef:
Closes #724

### DIFF
--- a/.github/workflows/cd.yml
+++ b/.github/workflows/cd.yml
@@ -68,7 +68,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        os: [ubuntu-16.04, macos-10.15, windows-2019]
+        os: [ubuntu-latest, macos-10.15, windows-2019]
     steps:
       - uses: actions/checkout@v2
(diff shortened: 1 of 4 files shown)


ppolewicz:
Has the binary size changed significantly? I'm thinking maybe we should have two releases, one dynamic and the other one static (and larger) for the weird environments. Have you tested it on `alpine`?

ppolewicz (approved):
Excellent, ready to merge

PR #752, opened 2021-09-23
Fix for static linking of Linux binary

### LINKED ISSUE #750: b2 version 3.0.2 won't run on CentOS 7.9 - libm.so.6 / GLIBC error
Hi

b2 version 3.0.2 won't run on CentOS version 7.9, giving the following error:

`[65850] Error loading Python lib '/root/tmp/_MEI1w886p/libpython3.9.so.1.0': dlopen: /lib64/libm.so.6: version `GLIBC_2.29' not found (required by /root/tmp/_MEI1w886p/libpython3.9.so.1.0)`

Possibly due to the old version of GLIBC that comes with CentOS 7.9?:

```
# ldd --version
ldd (GNU libc) 2.17
```

Version 3.0.1 runs fine.

For reference:
CentOS Linux release 7.9.2009
b2 is installed in  /root/.local/bin

Thanks

mlech-reef:
- Use python version from the container in CD
- Update the changelog

CD uses python container. See lined issue for details.

Closes #750

### DIFF
--- a/.github/workflows/cd.yml
+++ b/.github/workflows/cd.yml
@@ -5,17 +5,17 @@ on:
     tags: 'v*'  # push events to matching v*, i.e. v1.0, v20.15.10
 
 env:
-  PYTHON_DEFAULT_VERSION: 3.9
+  ACTIONS_STEP_DEBUG: ${{ secrets.ACTIONS_STEP_DEBUG }}
   GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+  PYTHON_DEFAULT_VERSION: 3.9
 
 jobs:
   deploy:
     env:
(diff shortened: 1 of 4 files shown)


.github/workflows/cd.yml
```diff
@@ -34,43 +34,59 @@ jobs:
         uses: mindsers/changelog-reader-action@v1
         with:
           version: ${{ steps.build.outputs.version }}
-      - name: Create GitHub release
+      - name: Create GitHub release and upload the distribution
         id: create-release
-        uses: actions/create-release@v1
```
mlech-reef, line 39: This is obsolete and was causing problems when run from the container. I use https://github.com/softprops/action-gh-release as in https://github.com/Backblaze/terraform-provider-b2

mszumocki-reef (approved):

PR #763, opened 2021-11-02
Do not install B2 SDK from git in CD

mlech-reef:


CHANGELOG.md
```diff
@@ -6,6 +6,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Fixed
+* Fix installing SDK in CD job
+
```
ppolewicz, line 11: I don't think this should be listed in the changelog - functionally the release is not changed in any way, it's just a little bit more resilient against people merging b2sdk branch to master during the cli release process. Not something we should notify the users about.

ppolewicz (approved):

PR #768, opened 2021-11-20
Fallback to ascii decoder in case locales are not properly set

### LINKED ISSUE #755: Use ascii codec for printing help when locales are not configured in the OS
This is mainly the problem when the binary is used with bare docker image like CentOS or Alpine:
image

### LINKED ISSUE #766: Ubuntu dependences?
Hi guys

Im trying to use the last version prom b2 CLI on Ubuntu 20:04. 

When I try to execute this b2-linux, I recive an error and this is the return.

```
  File "b2/console_tool.py", line 2448, in <module>
  File "b2/console_tool.py", line 2433, in main
  File "b2/console_tool.py", line 2224, in run_command
  File "argparse.py", line 1820, in parse_args
  File "argparse.py", line 1853, in parse_known_args
  File "argparse.py", line 2090, in _parse_known_args
  File "b2/arg_parser.py", line 80, in error
  File "argparse.py", line 2550, in print_help
  File "argparse.py", line 2556, in _print_message
UnicodeEncodeError: 'ascii' codec can't encode character '\u2022' in position 1084: ordinal not in range(128)
[478] Failed to execute script 'console_tool' due to unhandled exception!

```

Theres any dependence or mandatory package then I need to install?

mlech-reef:
- Fallback to ascii decoder in case locales are not properly set
- Update the changelog
- Format the code

Fixes #755
Fixes #766

### DIFF
--- a/b2/arg_parser.py
+++ b/b2/arg_parser.py
@@ -9,7 +9,9 @@
 ######################################################################
 
 import argparse
+import locale
 import re
+import sys
 import textwrap
 
 import arrow
@@ -68,7 +70,8 @@ class ArgumentParser(argparse.ArgumentParser):
             if self._for_docs:
                 self._description = textwrap.dedent(self._raw_description)
             else:
-                self._description = rst2ansi(self._raw_description.encode('utf-8'))
+                encoding = self._get_encoding()
+                self._description = rst2ansi(self._raw_description.encode(encoding), output_encoding=encoding)
 
         return self._description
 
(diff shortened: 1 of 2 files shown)


b2/arg_parser.py
```diff
@@ -78,9 +83,19 @@ def description(self, value):
...
         args = {'prog': self.prog, 'message': message}
         self.exit(2, '\n%(prog)s: error: %(message)s\n' % args)
 
+    @classmethod
+    def _get_encoding(cls):
+        _, locale_encoding = locale.getdefaultlocale()
+        if locale_encoding is not None:
+            return sys.getdefaultencoding()
+
+        # locales are improperly configured
```
ppolewicz, line 96: perhaps we should emit a warning in this case? If not, lets at least log it so that if someone runs in `--verbose` mode, they can clearly see what's going on

mlech-reef: No, it is only for printing help messages or argparse errors. All the parameters, including `--verbose` are not even evaluated in such a case, and the logger is disabled by deault so not one will see it. I could emit a python warning, but I don't think it is a valid case

PR #784, opened 2022-04-13
[DRAFT] Fix pyinstaller 4865

ppolewicz:
This will not really be merged from this PR, but instead it'll be released as a hotfix from a branch and then (if the fix is effective) it'll be cherrypicked into the main branch. Need the PR to run the tests though.


ppolewicz:
We are dropping macos binary in favor of `brew install b2-tools`

PR #800, opened 2022-06-17
Add replication-status command

agoncharov-reef:
This is the same as `replication-monitoring` branch but rebased on latest upstream master and added few commits (--noProgress and test_replication_monitoring).


ppolewicz (approved):
Needs to be pinned to a new version of b2-sdk-python

PR #838, opened 2022-11-17
Moving changelog action to the last working version

kkalinowski-reef:
Recent upgrade of dependencies of changelog changed the default behaviour
of unhandled promise rejection from just a warning to returning an error.
For a reason unknown, this happens only on master branch.

Issue was raised with the action creator, until it is resolved, an older
version of the action is used.


kkalinowski-reef:
Action owner confirmed that it should only be used with pull requests and it's not built for using with pushes. This is why the latest version failed on master. Better solution will be to disable changelog verification on pushes only.

PR #850, opened 2023-02-01
Prepare 3.7.0

ppolewicz:
run CI first


ppolewicz:
one thread safety fix still pending

PR #857, opened 2023-02-09
Prepare release 3.7.1

ppolewicz:


.github/workflows/ci.yml
```diff
@@ -147,7 +147,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        os: [macos-11, windows-2019]
+        os: [windows-2019]
```
kkalinowski-reef, line 150: Whole matrix can be dropped now, instead of:
```
    runs-on: ${{ matrix.os }}
    strategy:
      fail-fast: false
      matrix:
        os: [windows-2019]
```

you should be able to go with:
```
    runs-on: windows-2019
```

README.md
```diff
@@ -14,28 +14,40 @@ The latest documentation is available on [Read the Docs](https://b2-command-line
...
+
+Stand-alone binaries are available for Linux and Windows - this is the most straightforward way to use the command-line tool and is sufficient in most use cases. The latest versions are available for download from the [Releases page](https://github.com/Backblaze/B2_Command_Line_Tool/releases).
+
+## Python Package Index
+
+You can also install it in your python environment ([virtualenv](https://pypi.org/project/virtualenv/) is recommended) from PyPI with:
 
     pip install b2
 
+## Installing from source
```
kkalinowski-reef, line 29: How about adding a line about `brew` for `macOS`?

.github/workflows/cd.yml
```diff
@@ -83,54 +83,6 @@ jobs:
...
-        id: bundle
-        run: nox -vs bundle >> $GITHUB_OUTPUT
-      - name: Import certificate
-        uses: apple-actions/import-codesign-certs@v1
-        with:
-          p12-file-base64: ${{ env.B2_OSX_CODE_SIGNING_CERTIFICATE }}
-          p12-password: ${{ env.B2_OSX_CODE_SIGNING_CERTIFICATE_PASSWORD }}
-      - name: Sign the bundle
-        id: sign
-        run: nox -vs sign -- '${{ env.B2_OSX_CODE_SIGNING_CERTIFICATE_NAME }}' >> $GITHUB_OUTPUT
```
kkalinowski-reef, line 115: This means that we can remove a function that performs a signing for macOS from the `noxfile.py`.

.github/workflows/cd.yml
```diff
@@ -83,54 +83,6 @@ jobs:
...
           files: ${{ steps.sign.outputs.asset_path }}
-  deploy-macos-bundle:
-    needs: deploy
-    env:
-      B2_OSX_CODE_SIGNING_CERTIFICATE: ${{ secrets.B2_OSX_CODE_SIGNING_CERTIFICATE }}
-      B2_OSX_CODE_SIGNING_CERTIFICATE_NAME: ${{ secrets.B2_OSX_CODE_SIGNING_CERTIFICATE_NAME }}
-      B2_OSX_CODE_SIGNING_CERTIFICATE_PASSWORD: ${{ secrets.B2_OSX_CODE_SIGNING_CERTIFICATE_PASSWORD }}
-      B2_OSX_NOTARY_NAME: ${{ secrets.B2_OSX_NOTARY_NAME }}
-      B2_OSX_NOTARY_PASSWORD: ${{ secrets.B2_OSX_NOTARY_PASSWORD }}
-    runs-on: macos-10.15
```
kkalinowski-reef, line 94: OMG, this was running on macOS 10.15... Wow.

.github/workflows/ci.yml
kkalinowski-reef: Btw – shouldn't we consider `windows-latest` in here? Currently it'll be `windows-2022`.

ppolewicz: perhaps I should leave the matrix in and test on two windows versions?

README.md
ppolewicz: it's in the big doc I've linked, but I guess I could copy it

.github/workflows/cd.yml
ppolewicz: I'm a bit afraid that we'll need it back one day and a bit hopeful that we'll one day have "pyinstaller but better", but you are right, it should be removed

Issue #872, opened 2023-04-27
Fast rm sometimes fails unit tests


ppolewicz:
```
2023-04-26T22:47:22.2122674Z =================================== FAILURES ===================================
2023-04-26T22:47:22.2123147Z _____________________ TestRmConsoleTool.test_rm_fail_fast ______________________
2023-04-26T22:47:22.2124047Z [gw0] linux -- Python 3.7.13 /opt/hostedtoolcache/PyPy/3.7.13/x64/bin/python
2023-04-26T22:47:22.2124362Z 
2023-04-26T22:47:22.2124647Z self = <test.unit.test_console_tool.TestRmConsoleTool testMethod=test_rm_fail_fast>
2023-04-26T22:47:22.2124961Z 
2023-04-26T22:47:22.2125152Z     def test_rm_fail_fast(self):
2023-04-26T22:47:22.2125563Z         # Since we already have all the jobs submitted to another thread,
2023-04-26T22:47:22.2126008Z         # we can only rely on the log to tell when it stopped.
2023-04-26T22:47:22.2126487Z         expected_in_stdout = '''
2023-04-26T22:47:22.2127333Z         Deletion of file "b/b1/test.csv" (9996) failed: Conflict:
2023-04-26T22:47:22.2127797Z          count: 4/4'''
2023-04-26T22:47:22.2128219Z         unexpected_in_stdout = ' count: 5/5 '
2023-04-26T22:47:22.2128778Z >       self._run_problematic_removal(['--failFast'], expected_in_stdout, unexpected_in_stdout)
2023-04-26T22:47:22.2129103Z 
2023-04-26T22:47:22.2129305Z test/unit/test_console_tool.py:2661: 
2023-04-26T22:47:22.2131114Z _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
2023-04-26T22:47:22.2131553Z test/unit/test_console_tool.py:2651: in _run_problematic_removal
2023-04-26T22:47:22.2131988Z     unexpected_part_of_stdout=unexpected_in_stdout,
2023-04-26T22:47:22.2132394Z test/unit/test_console_tool.py:228: in _run_command
2023-04-26T22:47:22.2132787Z     self.assertIn(expected_part_of_stdout, actual_stdout)
[...]

PR #889, opened 2023-06-19
Fix Fast rm sometimes fails unit tests #872

przadka:
This pull request fixes race condition and reporter count update problems as reported in https://github.com/Backblaze/B2_Command_Line_Tool/issues/872.

The first fix is to ensure update_total is called before a task is submitted to the executor. This change was necessary to prevent a race condition in which the task could potentially finish before update_total is called, leading to unpredictable test outcomes. By ensuring that update_total is called before the task submission, we guarantee that the total count is always accurate at the time of task execution.

The second fix improves the updating of the reporter count. Initially, the reporter count was being updated before the future result was obtained, which could lead to inaccurate count values if the future result returned an exception. To make the tests more predictable and the reporter count more accurate, I modified the _removal_done function to only update the reporter's count after a successful future result. This ensures that the count is updated only when we are certain that the file has been deleted successfully. The test assert condition was updated to reflect this change.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,7 +14,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 * Update pyinstaller to fix Linux Bundle build
 
 ### Fixed
-* Fast rm sometimes fails unit tests
+* Fast rm sometimes failing due to a rare race condition
 
 ## [3.9.0] - 2023-04-28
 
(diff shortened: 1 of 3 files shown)


CHANGELOG.md
```diff
@@ -13,6 +13,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 * Increase timeout time in autocomplete tests to accommodate slower CI environments
 * Update pyinstaller to fix Linux Bundle build
 
+### Fixed
+* Fast rm sometimes fails unit tests
```
ppolewicz, line 17:
```suggestion
* Fast rm sometimes failing due to a rare race condition
```

ppolewicz (approved):

PR #899, opened 2023-07-27
Add ability to upload from an unbound source such as standard input or a named pipe

mjurbanski-reef:
It maybe best to use a squash merge here since there are so many commits here.


b2/console_tool.py
```diff
@@ -2598,43 +2608,227 @@ def run(self, args):
...
+            '--sha1', help="SHA-1 of the data being uploaded for verifying file integrity"
+        )
+        parser.add_argument(
+            '--threads', type=int, default=10, help="number of threads used for the operation"
+        )
+        parser.add_argument('--cache-control', default=None)
+        parser.add_argument(
+            '--info',
+            action='append',
+            default=[],
```
mpnowacki-reef, line 2647: the default being mutable can backfire in some strange corner case, no? @mjurbanski-reef

b2/console_tool.py
```diff
@@ -2598,43 +2608,227 @@ def run(self, args):
...
+            "file_name": args.b2FileName,
+            "file_retention": self._get_file_retention_setting(args),
+            "legal_hold": self._get_legal_hold_setting(args),
+            "local_file": args.localFilePath,
+            "min_part_size": args.minPartSize,
+            "progress_listener": make_progress_listener(args.localFilePath, args.noProgress),
+            "sha1_sum": args.sha1,
+        }
+
+    def execute_operation(self, **kwargs) -> 'b2sdk.file_version.FileVersion':
```
mpnowacki-reef, line 2710: why this and not abc?

b2/console_tool.py
```diff
@@ -2598,43 +2608,227 @@ def run(self, args):
...
+        if isinstance(file_id, (str, int)):
+            return open(
+                file_id,
+                mode="rb",
+                closefd=not isinstance(file_id, int),
+                buffering=kwargs.get('buffering', RECOMMENDED_READ_BUF_SIZE),
+            )
+        return file_id
+
+    class NotAnInputStream(Exception):
```
mpnowacki-reef, line 2744: I don't think I've seen his pattern in this project before, any reason for not declaring this exception at module level?

test/helpers.py
```diff
@@ -0,0 +1,19 @@
...
+#
+# License https://www.backblaze.com/using_b2_code.html
+#
+######################################################################
+import platform
+
+import pytest
+
+
+def skip_on_windows(*args, reason='Not supported on Windows', **kwargs):
```
mpnowacki-reef, line 15: this thing is already in master

b2/console_tool.py
mjurbanski-reef: because the only function rising it is a class method; i.e. if this function is removed, exception should be removed as well

We are talking about ~4k long file, which is why namespacing is especially important.

test/helpers.py
mjurbanski-reef: yes, it was just moved around in this PR

PR #903, opened 2023-08-02
Add retry_time parameter instead of hardcoded 5 retries

djaksic-reef:
Add retry_time parameter to retry for some period (in minutes) when an upload or download fails, instead of hardcoded 5 retries

https://reeftechnologies.atlassian.net/browse/B2-3


djaksic-reef:
Wrong place

PR #905, opened 2023-08-11
Updates `--corsRules` flag help text to specify what the command actually accepts

Shrinks99:
Minor help text addition, would have saved me some trial and error :)

### DIFF
--- a/b2/console_tool.py
+++ b/b2/console_tool.py
@@ -1043,7 +1043,12 @@ class CreateBucket(DefaultSseMixin, Command):
     @classmethod
     def _setup_parser(cls, parser):
         parser.add_argument('--bucketInfo', type=json.loads)
-        parser.add_argument('--corsRules', type=json.loads)
+        parser.add_argument(
+            '--corsRules',
+            type=json.loads
+            help=
+            "If given, the bucket will have a 'custom' CORS configuration. Accepts a JSON formatted string."
+        )
         parser.add_argument('--lifecycleRules', type=json.loads)
         parser.add_argument(
             '--fileLockEnabled',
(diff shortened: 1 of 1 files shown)


b2/console_tool.py
```diff
@@ -1043,7 +1043,12 @@ class CreateBucket(DefaultSseMixin, Command):
     @classmethod
     def _setup_parser(cls, parser):
         parser.add_argument('--bucketInfo', type=json.loads)
-        parser.add_argument('--corsRules', type=json.loads)
+        parser.add_argument(
+            '--corsRules',
+            type=json.loads
+            help=
+            "If given, the bucket will have a 'custom' CORS configuration. Accepts a JSON formatted string."
```
ppolewicz, line 1050:
```suggestion
            "If given, the bucket will have a 'custom' CORS configuration. Accepts a JSON string."
```
it doesn't need to be "formatted", right? I think it would be better to specify it as "JSON string", otherwise people will may think you have to do the `.indent()` thing on it before you pass it here.

Shrinks99: Works for me.  To resolve my confusion the important bit of information was that it's a JSON _string_ and not a path to a JSON _file_ :)

b2/console_tool.py
```diff
@@ -2539,7 +2544,12 @@ class UpdateBucket(DefaultSseMixin, Command):
     @classmethod
     def _setup_parser(cls, parser):
         parser.add_argument('--bucketInfo', type=json.loads)
-        parser.add_argument('--corsRules', type=json.loads)
+        parser.add_argument(
+            '--corsRules',
+            type=json.loads
```
ppolewicz, line 2549:
```suggestion
            type=json.loads,
```
fixing syntax error

b2/console_tool.py
```diff
@@ -1043,7 +1043,12 @@ class CreateBucket(DefaultSseMixin, Command):
...
```
ppolewicz, line 1048:
```suggestion
            type=json.loads
```
fixing syntax error

ppolewicz: @Shrinks99 you need to fix this one too

ppolewicz:
ok all is well now, but in order to let CI do its thing you have to change `CHANGELOG.md` adding an entry there such as
```
* Better help text for --corsRules
```

in the `### Changed` section

Shrinks99:
Changelog updated!

Sorry for the runaround for such a small change!  Thanks for walking me through your process, will be sure to update the changelog should I make any future PRs :)

ppolewicz:
Thanks for the contribution @Shrinks99 and sorry for the dribbling feedback. CI didn't point out the changelog entry until we cleared the syntax error...

We hope you come back with more help string updates if they bother you too much :)

PR #924, opened 2023-09-26
Aded --quiet parameter to download by file and id.

przadka:
Added ---quiet as a parameter to `download-file-by-id` and `download-file-by-name`.


mjurbanski-reef (changes requested):
thx;

Please also make it so that `--quiet` implies `--noProgress`. Basically, `--quiet` should make it so that nothing is emitted to stdout. This will make it consistent with tools such as wget and curl.
Then you can drop `--noProgress` from integration tests commands.

PR #927, opened 2023-10-03
fix `nox -s build` not gathering typeshed libs licenses properly

mjurbanski-reef:


.github/workflows/ci.yml
```diff
@@ -18,7 +18,7 @@ jobs:
           fetch-depth: 0
       - uses: codespell-project/actions-codespell@2391250ab05295bddd51e36a8c6295edb6343b0e
         with:
-          ignore_words_list: datas
+          ignore_words_list: datas re-use
```
kkalinowski-reef, line 21: How about excluding the LICENSE file entirely with:
```
with:
  exclude_file: src/foo
```
taken from https://github.com/codespell-project/actions-codespell

b2/console_tool.py
```diff
@@ -3454,6 +3454,8 @@ class License(Command):  # pragma: no cover
             'https://raw.githubusercontent.com/PythonCharmers/python-future/master/LICENSE.txt',
         'pefile':
             'https://raw.githubusercontent.com/erocarrera/pefile/master/LICENSE',
+        'https://github.com/python/typeshed':
```
kkalinowski-reef, line 3457: Are we using the repository version? Why? Even their original readme says that you don't need to and points to:
```
pip install types-six types-requests
```
instead. Those these libraries provide no licensing information?
https://github.com/python/typeshed#using

kkalinowski-reef (approved):

.github/workflows/ci.yml
mjurbanski-reef: it also errored out in other files on the same word

b2/console_tool.py
mjurbanski-reef: 1) no, we aren't. This is URL extracted by pip-licenses. In fact this is not even a direct dependency of ours, so we don't specify it anywhere ourselves.
2) piplicense is unable to get LICENSE text for it. It is able to extract just license type (how it does it, idk, but I expect it to be in at least python trove classifiers for this package). It is not enough for our purpose, hence the direct link.

PR #965, opened 2023-11-24
Cache argument parser for quicker autocomplete suggestions

vbaltrusaitis-reef:


test/unit/_cli/test_autocomplete_cache.py
```diff
@@ -0,0 +1,289 @@
...
+
+import argcomplete
+import pytest
+
+import b2._cli.argcompleters
+import b2.arg_parser
+import b2.console_tool
+from b2._cli import autocomplete_cache
+
+from ..console_tool.conftest import *  # noqa
```
mjurbanski-reef, line 34: this seems entirely unnecessary; i.e. it should work exactly the same without this

PR #993, opened 2024-02-12
Don't show full binary path in help message

dawidd6:
Without this change:

```sh
root@dc7d16c9b121:/usr/local/lib/python3.11/site-packages/b2# b2 --help
/usr/local/bin/b2 [-h] [--help-all]
```

With this change:

```sh
root@dc7d16c9b121:/usr/local/lib/python3.11/site-packages/b2# b2 --help
b2 [-h] [--help-all]
```

### DIFF
--- a/b2/_internal/console_tool.py
+++ b/b2/_internal/console_tool.py
@@ -4006,7 +4006,7 @@ class ConsoleTool:
 
     def run_command(self, argv):
         signal.signal(signal.SIGINT, keyboard_interrupt_handler)
-        parser = B2.create_parser(name=argv[0])
+        parser = B2.create_parser(name=NAME)
         AUTOCOMPLETE.cache_and_autocomplete(parser)
         args = parser.parse_args(argv[1:])
         self._setup_logging(args, argv)


mjurbanski-reef:
This is a much-needed fix, but I'm afraid the current form is not going to cut it.

We recently introduced what we call apiver aliases, i.e.  `b2v3`  for b2 v3 stable API (no matter which version of b2 binary you installed), and `_b2v4` for yet unreleased v4 (after v4 release it will become another `b2v4` to be used in scripts etc.

Cases that should be considered:
1) when calling by `b2` we expect `b2` to be shown in --help text
2) when calling by `b2v3` name we expect `b2v3` to be shown in --help text
3) ideally if someone renamed binary we would expect that name to show up in `--help`, but this maybe problematic to achieve in practice. While not recommened, someone could run this as python script i.e. `python .../b2.py` and in that case we wouldn't like to mistakenly show `python` as "b2" exec in --help (this bug is present right now).

1 and 2 cases should be already tested and if aren't, then we need to add tests covering them.

ppolewicz:
@dawidd6 would you like to do the extra changes or have someone take over?

dawidd6:
Someone can take this over. Tbh I don't understand what needs to be done in this PR.

mjurbanski-reef:
resolved in #1002

Thank you for brining our attention to this issue.

PR #1000, opened 2024-03-13
Unify CLI argument flag style

vbaltrusaitis-reef:
Both style flags are still supported in CLI v3 binary (although using `--camelCase` will result in warnings). v4 will only support `--dash-style`.


ppolewicz:
Some users monitor stderr and if we start emitting new warnings in v3,
their alerts will go off. Can --camelCase remain to be allowed in b2v3?

mjurbanski-reef:
@ppolewicz we already made this decision to emit deprecation warnings to stderr when we deprecated multiple popular commands like `download-file-by-name` in version https://github.com/Backblaze/B2_Command_Line_Tool/releases/tag/v3.14.0 . In 3 months since this change, we have not received a single complaint.

The motivation to do it this way, was to "ease" people into new version.
IF we ever receive such complaints we may introduce an ENV var silencing deprecation warnings, but still show them by default.

PR #1018, opened 2024-04-23
fix GHA CI by using old macos for old python versions

mjurbanski-reef:
Workaround for https://github.com/actions/setup-python/issues/696


mjurbanski-reef:
that is what was used before macos-14 rollout under macos-latest name, but I guess macos-13 does come with 4 CPU cores instead of 3 and has longer support  https://github.com/actions/runner-images/issues/9255 , so I will change to macos-13 as suggested.

PR #1054, opened 2024-10-31
upgrade to pytest 8

selfisekai:


ppolewicz:
Hey, are you going to move it forward from draft?

selfisekai:
i put it as draft one test always fails, but i'm afraid it's not getting fixed https://github.com/pytest-dev/pytest-forked/issues/67

mlech-reef:
@selfisekai, but the CI passed - even `test_autocomplete_cache.py`. Where exactly you see the problem?

selfisekai:
interesting, it does pass for me locally on `pdm run nox -s unit-3.12`, but not on `pdm run pytest -k 'not integration'`:
```
self = <_pytest.runner.SetupState object at 0x7f06f7f698e0>, item = <Function test_add_or_update_shell_section_new_section>

    def setup(self, item: Item) -> None:
        """Setup objects along the collector chain to the item."""
        needed_collectors = item.listchain()
    
        # If a collector fails its setup, fail its entire subtree of items.
        # The setup is not retried for each item - the same exception is used.
        for col, (finalizers, exc) in self.stack.items():
>           assert col in needed_collectors, "previous item was not torn down properly"
E           AssertionError: previous item was not torn down properly

.venv/lib/python3.12/site-packages/_pytest/runner.py:505: AssertionError
```

mlech-reef:
@selfisekai the problem is order of tests in `test_autocomplete_cache.py`. It is a mix of forked and non-forked tests, where non-forked are not the last ones in the file. And this is a problem according to [pytest-dev/pytest-forked/issues/67](https://github.com/pytest-dev/pytest-forked/issues/67#issuecomment-1964718720)

I propose to move `test_pickle_store` and `test_unpickle` to the separate module, e.g. called `test/unit/_cli/test_pickle.py`. And then, everything should be fine.

BR/ Maciek

ppolewicz:
@selfisekai 3.8 tests fail, from the logs it looks like it's something that should be easy to fix (type annotation?)

PR #1062, opened 2024-12-13
Fix shell autocompletion for Python >=3.12.8 and >=3.13.1

mlech-reef:


mlech-reef:
CI failures not related to the changes, caused by the broken `master` branch - they are going to be fixed in #1060.

olzhasar-reef (approved):

PR #1102, opened 2025-07-30
Pin docutils version

olzhasar-reef:
Some parts of the docutils were deprecated in 0.22.x that
the dependent rst2ansi package relied upon


mlech-reef (approved):

olzhasar-reef:
@mlech-reef 
It is indeed dead.
This is a temporary solution to ensure the app is functioning, but we'd need to find a proper replacement for `rst2ansi`.

PR #1108, opened 2025-08-03
Add Linux binary for ARM

mlech-reef:


.github/workflows/cd.yml
```diff
@@ -22,8 +22,9 @@ jobs:
     steps:
       - name: Determine if pre-release
         id: prerelease_check
+        # it checks for PEP 440 compatible prerelease tags
         run: |
-          export IS_PRERELEASE=$([[ ${{ github.ref }} =~ [^0-9]$ ]] && echo true || echo false)
+          export IS_PRERELEASE=$([[ ${{ github.ref }} =~ [abrcdev]+[0-9]+$ ]] && echo true || echo false)
```
olzhasar-reef, line 27: Shouldn't it be something like this:
```
(a|b|rc|dev)[0-9]+$
```

olzhasar-reef (approved):

PR #1144, opened 2026-04-29
Drop python3.9 support + upgrade dependencies

olzhasar-reef:
- **Bump pyinstaller version, remove obsolete hidden_imports from spec file**
- **Drop Python 3.9 and PyPy 3.9 support**
- **Rewrite the test case veryifying non-UTF-8 help rendering to avoid subtests**
- **Bump test dependencies versions**


test/unit/test_arg_parser.py
olzhasar-reef, line 1: Had to rewrite the existing test case, because the newer version of pytest had some issues with serializing classes when using subTests. Going away from subTests seems like a sane thing to do in any case

.github/workflows/ci.yml
```diff
@@ -103,7 +103,7 @@ jobs:
       fail-fast: false
       matrix:
         os: ["ubuntu-latest", "ubuntu-24.04-arm", "macos-latest", "windows-latest"]
-        python-version: ["3.9", "3.10", "3.11", "3.12", "3.13", "3.14", "pypy3.9", "pypy3.10"]
```
olzhasar-reef, line 106: I doubt that there's much value in testing against two pypy versions here, so I removed one assuming the tail one should be enough.

mlech-reef (approved):

